PR #5, opened 2018-11-07
Support AMI built from non-Amazon linux distributions

zaro0508:
The managed-ec2 CF template expects a success signal after spinning up
an instance.  The problem is that non amazon linux distrubtions do not
send the success signal so the cloudformation deployment would just
hang, time out then fail.  The workaround is to listen for the
success signal when deploying non amazon linux distrubtions.

The side affect of this workaround:
1. The auto provisioner will spin up a non amazon linux instance and tell
   you that it’s available before the EC2 gets completely spun up.
2. The system may tell you that an EC2 is spun up but it may actually
   fail to spin up successfully.

### DIFF
--- a/templates/managed-ec2.yaml
+++ b/templates/managed-ec2.yaml
@@ -111,9 +111,26 @@ Parameters:
     Description: ID of the AMI to deploy
     Type: String
     Default: ""
+  AMILinuxDist:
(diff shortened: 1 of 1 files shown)


zaro0508:
This is old and has been figured out in existing windows and ubuntu EC2 templates

PR #8, opened 2018-11-07
added new parameter to restrict download/upload to only AWS resources (PLFM-5194)

zimingd:


templates/SynapseExternalBucket.yaml
```diff
@@ -123,6 +136,111 @@ Resources:
...
+            #generate new policy statement based on data from AWS
+            ip_ranges = requests.get('https://ip-ranges.amazonaws.com/ip-ranges.json').json()['prefixes']
+            region_ip_addresses = [item['ip_prefix'] for item in ip_ranges if (item["service"] == "AMAZON" and item["region"] == region)]
+            
+            new_ip_policy_statement = {'Sid': policy_statement_id,
+                                       'Effect': 'Deny',
+                                       'Principal': '*',
+                                       'Action': 's3:GetObject',
+                                       'Resource': 'arn:aws:s3:::'+bucket_name+'/*',
+                                       'Condition': {'NotIpAddress': {'aws:SourceIp': region_ip_addresses}}}
```
zaro0508: it looks like 'aws:SourceIP' need to be a list.  https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_examples_aws_deny-ip.html

Aren't you missing the brackets '[]' for list notation and commas between IP addresses?

templates/SynapseExternalBucket.yaml
```diff
@@ -123,6 +136,111 @@ Resources:
...
+            bucket_policy = json.loads(s3_client.get_bucket_policy(Bucket=bucket_name)['Policy'])
+
+            #remove the old IP address policy statement and add new
+            bucket_policy['Statement'] = [statement for statement in bucket_policy['Statement'] if (policy_statement_id != statement.get("Sid"))]
+            bucket_policy['Statement'].append(new_ip_policy_statement)
+
+
+            s3_client.put_bucket_policy(Bucket=bucket_name, Policy=json.dumps(bucket_policy))
+
+        - { BucketName: !If [EnableEncryption, !Ref SynapseEncryptedExternalBucket, !Ref SynapseExternalBucket], Region: !Ref "AWS::Region" }
```
zaro0508: This doesn't look like the right notation for passing in variables.  https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-environment.html

shouldn't it be something like this?
https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/RotateCredentials.yaml#L47

templates/SynapseExternalBucket.yaml
```diff
@@ -123,6 +136,111 @@ Resources:
...
+          Version: '2012-10-17'
+          Statement:
+          - Effect: Allow
+            Action:
+            - logs:CreateLogGroup
+            - logs:CreateLogStream
+            - logs:PutLogEvents
+            Resource: "*"
+          - Action:
+            - s3:*
```
zaro0508: Shouldn't we scope the action to only allow updating bucket policies?

templates/SynapseExternalBucket.yaml
```diff
@@ -123,6 +136,111 @@ Resources:
...
+    Properties:
+      AssumeRolePolicyDocument:
+        Version: '2012-10-17'
+        Statement:
+        - Effect: Allow
+          Principal:
+            Service: lambda.amazonaws.com
+          Action: sts:AssumeRole
+      Path: "/"
+      Policies:
```
zaro0508: Should add an additional policy to allow lambda to publish to cloudwatch.

```
        - PolicyName: PublishToCloudwatch
          PolicyDocument:
            Version: 2012-10-17
            Statement:
              - Effect: Allow
                Action:
                  - 'logs:CreateLogGroup'
                  - 'logs:CreateLogStream'
                  - 'logs:PutLogEvents'
                Resource: 'arn:aws:logs:*:*:*'
```

templates/SynapseExternalBucket.yaml
```diff
@@ -123,6 +136,111 @@ Resources:
...
+      - PolicyName: external-bucket-lambda-policy
+        PolicyDocument:
+          Version: '2012-10-17'
+          Statement:
+          - Effect: Allow
+            Action:
+            - logs:CreateLogGroup
```
zaro0508: per my suggestion above.  setup separate policies for each resource.

templates/SynapseExternalBucket.yaml
```diff
@@ -123,6 +136,111 @@ Resources:
...
+      Endpoint: !GetAtt ExternalBucketGroupPolicyUpdateLambda.Arn
+  ExternalBucketGroupPolicyUpdateLambdaSNSInvokePermission: # for sns to trigger group policy update lambda
+    Condition: CreateIPAddressRestrictionLambda
+    Type: AWS::Lambda::Permission
+    Properties:
+      FunctionName: !GetAtt ExternalBucketGroupPolicyUpdateLambda.Arn
+      Action: 'lambda:InvokeFunction'
+      SourceArn: 'arn:aws:sns:us-east-1:806199016981:AmazonIpSpaceChanged'
+      Principal: sns.amazonaws.com
+  ExternalBucketGroupPolicyUpdateTimer:
```
zaro0508: why do you need 2 separate triggers for updating?

templates/SynapseExternalBucket.yaml
```diff
@@ -25,9 +35,11 @@ Parameters:
   SynapseUserName:
     Type: String
     Description: Synapse user for read-write bucket
+    AllowedPattern: ".+" # makes this parameter required
```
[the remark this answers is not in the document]
zimingd: When left empty on the clouformation console webpage, errors would occur after the resources were created and the Outputs were being processed made. Users are not allowed to proceed these parameters were not empty

templates/SynapseExternalBucket.yaml
```diff
@@ -123,6 +136,111 @@ Resources:
...
+      Timeout: 10
+      Code:
+        ZipFile: !Sub
+        - |
+          import json
+          import boto3
+          from botocore.vendored import requests
+
+          bucket_name = '${BucketName}'
+          region = '${Region}'
```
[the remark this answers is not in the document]
zimingd: It would only work on us-east-1. Bruce said this would be fine until we decided to support other regions in the backend.

templates/SynapseExternalBucket.yaml
zimingd: region_ip_addresses is already a list type. The brackets are automatically added by the json library.

templates/SynapseExternalBucket.yaml
zimingd: Yeah the SNS trigger should be enough for updates.
I added a new CloudFormation Custom Resource to make the lambda apply the group policy when during the execution of the template.

zimingd:
@zaro0508 review changes have been made

templates/SynapseExternalBucket.yaml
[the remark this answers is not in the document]
zaro0508: hmm, interesting.  Cloudformation would run and resources would get created even if you don't set  a value for SyanpaseUserName?  That's not my experience with the CLI.  it will fail without running the template at all.  This system uses the CLI

templates/SynapseExternalBucket.yaml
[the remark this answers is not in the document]
zaro0508: That's fine but could we add some comments here to let people know.  Also might be nice to include docs in the README as well.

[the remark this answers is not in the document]
zaro0508: Also maybe add doc to the parameter description?

templates/SynapseExternalBucket.yaml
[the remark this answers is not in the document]
zimingd: I'll remove the regex then. I thought users might be taking this template and running it on the aws console.

PR #20, opened 2018-11-14
Allow access to public EC2 instances with an SSH keypair

zaro0508:
Allow SSH access to public and private EC2 instances.  This will allow
users to access public instances from with an SSH keypair.
This does not affect private instances because they are still deployed
to a private subnet.  Private instances are still only accessible
after logging into the Sage VPN.


zaro0508:
replacing this one with https://github.com/Sage-Bionetworks/scicomp-provisioner/pull/24

PR #25, opened 2018-11-16
Test base ubuntu image

thomasyu888:


thomasyu888:
This AMI is 8GB, built from ubuntu distribution.

config/prod/test-ubuntu.yaml
```diff
@@ -0,0 +1,21 @@
+# Provision an EC2 instance
+template_path: templates/managed-ec2.yaml
```
zaro0508: `managed-ec2.yaml` has been deprecated.  Please change this to reference the new template version `managed-ec2-v2.yaml`

config/prod/test-ubuntu.yaml
```diff
@@ -0,0 +1,21 @@
...
+parameters:
+  # The Sage deparment for this resource
+  Department: "CompOnc"
+  # The Sage project this resource will be used for
+  Project: "Ubuntu base image"
+  # The resource owner
+  OwnerEmail: "thomas.yu@sagebase.org"
+  # EC2 instance type (available types https://aws.amazon.com/ec2/instance-types/)
+  InstanceType: "t2.micro"
+  AMIId: "ami-010d7285c460bd379"
```
zaro0508: Actually this is not going to work with managed-ec2-v2.yaml template because there's custom commands in that template to install stuff using the yum package manager.  https://github.com/Sage-Bionetworks/scicomp-provisioner/blob/master/templates/managed-ec2-v2.yaml#L500

I think to get this working we'll need to update the template to support distributions that use apt package manager.

thomasyu888: Are we going to work on allowing for ubuntu?  I want to push for the allowance of Amazon Linux and Ubuntu.  Theres not too much difference between linux and ubuntu, but I know personally I use Ubuntu over Linux. (honestly its just a difference between installation of dependencies)

zaro0508: We will try.  I've already worked on it for a day but was not able to get it working.  The errors I ran into were are attached to https://sagebionetworks.jira.com/browse/IT-391

thomasyu888:
Closing for now.

PR #26, opened 2018-11-16
Test amazon linux 2

thomasyu888:


thomasyu888:
This AMI is 8GB, built from amazon linux 2 distribution.

thomasyu888:
I should probably update the AMI's for amazon linux distribution to use amazon linux 2.  

https://serverfault.com/questions/889248/how-to-enable-systemd-on-amazon-linux-ami
https://aws.amazon.com/amazon-linux-2/faqs/

This goes back to the issue of, I don't think we want to be having to maintain updates to AMI's all the time when new AMI's come out.  If there was a way for the template to spin up an ec2 with encrypted bootdisk without an AMI, that would be amazing.

config/prod/test-amazonlinux2.yaml
```diff
@@ -0,0 +1,21 @@
+# Provision an EC2 instance
+template_path: templates/managed-ec2.yaml
```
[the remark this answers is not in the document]
thomasyu888: What are the differences between templates?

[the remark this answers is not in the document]
zaro0508: if we update the current yaml CF will terminate the existing instance and deploy new ones which will cause a lot of pain for people who are using existing instances.  

reason for change to v2: https://github.com/Sage-Bionetworks/scicomp-provisioner/pull/24

zaro0508:
Do we still need this test?

thomasyu888:
I'm thinking about it.  The default distribution is the Amazon Linux 1.  I wonder if we should update.  Keep this open for now.

zaro0508:
We have it defaulted to `ami-0ddee041772c2d9f8` isn't the source of this image from amazon linux 2 AMI?

thomasyu888:
No this is the amazon linux 1.

zaro0508:
ahh ok.  then yes, we should definitely use AWS linux 2

config/prod/test-amazonlinux2.yaml
```diff
@@ -0,0 +1,21 @@
...
+parameters:
+  # The Sage deparment for this resource
+  Department: "CompOnc"
+  # The Sage project this resource will be used for
+  Project: "Amazon linux 2 base image"
+  # The resource owner
+  OwnerEmail: "thomas.yu@sagebase.org"
+  # EC2 instance type (available types https://aws.amazon.com/ec2/instance-types/)
+  InstanceType: "t2.micro"
+  AMIId: "ami-056131b84aa59cae0"
```
zaro0508: this AMI's disk is NOT encrypted, /dev/xvda=snap-0696504305c0108ce:8:true:gp2

lets test an AMI with an encrypted disk.

config/prod/test-amazonlinux2.yaml
```diff
@@ -0,0 +1,21 @@
+# Provision an EC2 instance
+template_path: templates/managed-ec2-v2.yaml
```
zaro0508: use managed-ec2-v3.yaml

config/prod/test-amazonlinux2.yaml
zaro0508: Test with VolumeSize parameter as well?

thomasyu888:
Will make all these changes soon.

PR #27, opened 2018-11-16
Test redhat linux

thomasyu888:
This AMI is 8GB built from redhat linux instance

### DIFF
--- a/config/prod/test-redhatlinux.yaml
+++ b/config/prod/test-redhatlinux.yaml
@@ -1,11 +1,11 @@
 # Provision an EC2 instance
-template_path: templates/managed-ec2.yaml
+template_path: templates/managed-ec2-v2.yaml
 stack_name: amazon-linux2
 parameters:
   # The Sage deparment for this resource
   Department: "CompOnc"
   # The Sage project this resource will be used for
-  Project: "Amazon linux 2 base image"
+  Project: "Redhat linux base image"
   # The resource owner
   OwnerEmail: "thomas.yu@sagebase.org"
   # EC2 instance type (available types https://aws.amazon.com/ec2/instance-types/)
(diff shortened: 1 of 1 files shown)


config/prod/test-redhatlinux.yaml
```diff
@@ -0,0 +1,21 @@
+# Provision an EC2 instance
+template_path: templates/managed-ec2.yaml
+stack_name: amazon-linux2
+parameters:
+  # The Sage deparment for this resource
+  Department: "CompOnc"
+  # The Sage project this resource will be used for
+  Project: "Amazon linux 2 base image"
```
zaro0508: is this amazon linux or redhat?

config/prod/test-redhatlinux.yaml
```diff
@@ -0,0 +1,21 @@
...
+  # The resource owner
+  OwnerEmail: "thomas.yu@sagebase.org"
+  # EC2 instance type (available types https://aws.amazon.com/ec2/instance-types/)
+  InstanceType: "t2.micro"
+  AMIId: "ami-02219c267315128db"
```
zaro0508: i don't think redhat will work with `managed-ec2-v2.yaml` template due to installation of aws-cfn-bootstrap package.  https://github.com/Sage-Bionetworks/scicomp-provisioner/blob/master/templates/managed-ec2-v2.yaml#L500


it seems like it's slightly different to install on redhat.  https://forums.aws.amazon.com/thread.jspa?threadID=226199

thomasyu888: Hmmm... I don't use this distribution of linux ever, but just thought I would try it out.  I think if we just limited people to work with the amazon linux distributions / ubuntu, we don't need to include the redhat linux.  Thoughts?

zaro0508:
Hmm, I think maybe we should setup a `sanbox-provisioner` for this type of testing?  But we don't have it yet, so give this a try..

zaro0508:
this test didn't work.  the template failed to install aws-cfn-bootstrap package

PR #29, opened 2018-11-20
add separate template for restricting same region download from bucket

zimingd:


zimingd:
Moved to 
https://github.com/Sage-Bionetworks/aws-infra/pull/69

PR #33, opened 2018-12-06
new AMI for tesla analysis

xdoan:


config/prod/tesla-analysis.yaml
```diff
@@ -0,0 +1,23 @@
+# Provision an EC2 instance
+template_path: templates/managed-ec2-v2.yaml
+stack_name: tesla-analysis
```
zaro0508: would you mind adding `ec2` the this name?  and also change the filename to match?  something like `tesla-analysis-ec2` would work.

config/prod/tesla-analysis.yaml
```diff
@@ -0,0 +1,23 @@
...
+parameters:
+  # The Sage deparment for this resource
+  Department: "CompOnc"
+  # The Sage project this resource will be used for
+  Project: "TESLA"
+  # The resource owner
+  OwnerEmail: "xengie.doan@sagebase.org"
+  # EC2 instance type (available types https://aws.amazon.com/ec2/instance-types/)
+  InstanceType: "m5.2xlarge"
```
zaro0508: wondering why you chose m5 instance?  why not t3 or t2?

xdoan: Tom and I were browsing through resources and STAR aligner recommended 100G disk/30G RAM/6+ CPUs. I would be happy to change it if t3 or t2 works better.

xdoan: Probably do not need that much disk space if I run through batches of files and delete them, but anything with that much memory and CPUs is good.

zaro0508: all have the same specs[1],  16 GB ram and 4 vCPU.  if you want something larger then i would recommend t3.2xlarge ?  AWS says all should work the same except t3 is the least expensive option.

[1] https://aws.amazon.com/ec2/instance-types/

xdoan: Sounds good, I changed it to t3.2xlarge.

PR #39, opened 2018-12-15
[IT-385] Dynamically set EC2 volume size

zaro0508:
Add a parameter to allow users to dynamically set the EC2
disk size when it gets provisioned.


templates/managed-ec2-v3.yaml
```diff
@@ -124,6 +124,12 @@ Parameters:
     Description: ID of the AMI to deploy
     Type: String
     Default: ami-0ddee041772c2d9f8
+  VolumeSize:
+    Description: The EC2 volume size (in GB)
+    Type: Number
+    Default: 8
+    MinValue: 8
+    MaxValue: 100
```
thomasyu888: There is already a use case where @xdoan needs 500GB for analysis.  Can we push up the max value?

PR #52, opened 2019-01-22
add templates to provision S3 static websites

zaro0508:


templates/managed-s3WebCloudfront.yaml
```diff
@@ -0,0 +1,145 @@
...
+        Enabled: true
+        HttpVersion: 'http2'
+        DefaultRootObject: index.html
+        Aliases:
+          - !Join
+            - '.'
+            - - !Ref SubDomainName
+              - !Ref DomainName
+        DefaultCacheBehavior:
+          AllowedMethods:
```
xschildw: We should add DefaultTTL: 3600 so we check every hour against the origin when a file is requested.

xschildw: I meant _above_ AllowedMethods of course...

zaro0508: done

PR #61, opened 2019-02-08
Request EC2 instance.

xschildw:


xschildw:
To host a Shiny server for the challenge. Will need to add a public application load balancer in front of it.

PR #64, opened 2019-02-19
new mpnst analysis with more resources

xdoan:


config/prod/mpnst-analysis-large.yaml
```diff
@@ -0,0 +1,21 @@
+# Provision an EC2 instance
+template_path: managed-ec2.yaml
```
zaro0508: a new version is available. please change to `template_path: managed-ec2-v4.yaml`

PR #65, opened 2019-02-20
Option to grant access to bucket

zaro0508:
This is to setup cross account access[1]. Add a parameter to
allow the provisioner to specify which account, group and user
can have access to the provisioned bucket. It is not a required
parameter, if nothing is specified then no access will be given.

[1] https://docs.aws.amazon.com/AmazonS3/latest/dev/example-walkthroughs-managing-access-example2.html

### DIFF
--- a/config/prod/amp-emory-bucket.yaml
+++ b/config/prod/amp-emory-bucket.yaml
@@ -1,7 +1,11 @@
 # Provision a Synapse External Bucket (http://docs.synapse.org/articles/custom_storage_location.html)
-template_path: SynapseExternalBucket.yaml
+template_path: SynapseExternalBucket-v2.yaml
 stack_name: AMP-Emory-Bucket
 parameters:
+  # The Sage deparment for this resource
+  Department: "SysBio"
+  # The Sage project this resource will be used for
+  Project: "AMP-Emory"
   # true for read-write bucket, false (default) for read-only bucket
   AllowWriteBucket: "true"
   # Synapse username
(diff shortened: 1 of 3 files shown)


config/prod/gates-ki-001-s3bucket.yaml
```diff
@@ -1,15 +1,22 @@
 # Provision a Synapse External Bucket (http://docs.synapse.org/articles/custom_storage_location.html)
-template_path: SynapseExternalBucket.yaml
+template_path: SynapseExternalBucket-v2.yaml
 stack_name: gates-ki-001
 parameters:
+  # The Sage deparment for this resource
+  Department: "Computational Oncology"
```
kdaily: This is not Comp Onc. This is Sys bio

kdaily:
```suggestion
  Department: "SysBio"
```

config/prod/amp-emory-bucket.yaml
```diff
@@ -1,15 +1,24 @@
 # Provision a Synapse External Bucket (http://docs.synapse.org/articles/custom_storage_location.html)
-template_path: SynapseExternalBucket.yaml
+template_path: SynapseExternalBucket-v2.yaml
 stack_name: AMP-Emory-Bucket
 parameters:
+  # The Sage deparment for this resource
+  Department: "Computational Oncology"
```
kdaily:
```suggestion
  Department: "SysBio"
```

PR #3, opened 2018-11-06
Example PR: Auto provision an EC2 Linux instance

zaro0508:
This is an example PR (do not merge!). It is a template to show
how to get the CI system to automatically provision an EC2 instance.

Instance creation workflow:

* Scientist create a PR similar to this one. Create a new file
* with a unique stack_name and updated parameters.
* Have someone review and approve the PR.
* Merge the PR
* The CI system will auto provision an EC2 instance
* An email will be sent to the owner email address with the new bucket info.

The newly created instance will contain the following tags:
```
key: Department
value: Platform

key: OwnerEmail
value: joe.smith@sagebase.org

key: Project
value: Infrastructure

key: aws:cloudformation:logical-id
value: Ec2Instance

key: aws:cloudformation:stack-id
value: arn:aws:cloudformation:us-east-1::stack/kickstart-ec2-service/559e55a0-bc47-11e8-995a-500c28b4b099

key: aws:cloudformation:stack-name
value: iwanna-instance
```

To remove the instance:

* Instance owner must create a new PR that deletes the files that was added by
the PR that was used to add the instance.
* Have an an AWS account Admin review your change.
* Approve and merge.

PR #73, opened 2019-03-07
updated to run on cNF data

sgosline:
is this how i get an ec2 instance?


zaro0508:
This is only to change the ownership of the existing EC2 but won't create a new EC2 for you.  If you want a new EC2 for yourself then follow instructions here: https://github.com/Sage-Bionetworks/scicomp-provisioner/pull/3

PR #75, opened 2019-03-07
trying again to get an instance!

sgosline:


config/prod/nf-salmon.yaml
```diff
@@ -1,8 +1,6 @@
-
-    
 # Provision an EC2 instance
 template_path: managed-ec2-v5.yaml
-stack_name: test-scrna
+stack_name: salmon-rnaSeq
```
zaro0508: not absolutely necessary but would help if the stack_name is the same as the file name or vise versa.

sgosline: fixed

PR #80, opened 2019-03-11
New neurolincs instance

philerooski:
I ended up using the original neurolincs instance #60 for something else (that still needs to run for the next few weeks).


philerooski:
Closing because I `git mv`'d neurolincs.yaml when I should've `cp`'d ...

PR #81, opened 2019-03-11
New neurolincs instance

philerooski:
Replaces #80

### DIFF
--- a/config/prod/neurolincs-actual.yaml
+++ b/config/prod/neurolincs-actual.yaml
@@ -1,5 +1,5 @@
 # Provision an EC2 instance
-template_path: managed-ec2-v4.yaml
+template_path: managed-ec2-v5.yaml
 stack_name: neurolincs-actual
 parameters:
   # The Sage deparment (Platform, CompOnc, SysBio, Governance, etc..)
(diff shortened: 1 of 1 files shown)


config/prod/neurolincs-actual.yaml
```diff
@@ -0,0 +1,22 @@
+# Provision an EC2 instance
+template_path: managed-ec2-v4.yaml
```
thomasyu888: Use new template

philerooski: Is v4 no longer supported?

thomasyu888: I don't think its not supported, but you always want to use the newer versions as they become available.

philerooski: I understand, but v4 works for my needs. If you no longer want people to open PRs using the v4 or v3 templates you should deprecate and remove them from the templates directory. I'll be happy to reopen this PR using v5 if you decide to do so.

zaro0508: both of you are correct.  I forgot to deprecate managed-ec2-v4.yaml.  It is deprecated now. https://github.com/Sage-Bionetworks/scicomp-provisioner/blob/master/templates/managed-ec2-v4.yaml

philerooski:
@zaro0508 @thomasyu888 changes made

PR #82, opened 2019-03-13
have to run more files, need more RAM and disk space

xdoan:


config/prod/mpnst-analysis-larger.yaml
```diff
@@ -0,0 +1,22 @@
+# Provision an EC2 instance
+template_path: managed-ec2-v4.yaml
```
thomasyu888: Use new template version

PR #100, opened 2019-03-26
Create usage-statistics-s3.yaml

kdaily:
I need to have non-Synapse access to this bucket. I'm not sure which IAM ARN to add under `GrantAccess`. Please advise!


kdaily:
I was looking at an example in the repo and there were some user-specific IAMs in there to grant access. Didn't know if I needed to specify.

https://github.com/Sage-Bionetworks/scicomp-provisioner/blob/master/config/prod/JKG-SciComp-S3.yaml#L22

PR #123, opened 2019-04-12
setup a bit and fast ebs volume for wpoehlm-ios-ec2 instance

zaro0508:


zaro0508:
volumes cannot be shutdown therefore we will be paying as long as it's been created on our accounts. https://aws.amazon.com/premiumsupport/knowledge-center/ebs-volume-charges/

wpoehlm:
> 
> 
> Is there not a limit on IOPS depending on size (i.e. can we get 64K with 1TB volume)?

looks like 64k is the maximum per volume.  I guess if we added a second volume we could reach the max iops per instance (80k), but that doesn't sound cost effective.  

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSVolumeTypes.html

By the way @zaro0508 , I am meeting with Ben this afternoon to finalize this plan.  I'll let you know when I'm ready for merging.

config/prod/wpoehlm-largevol-ec2-ebs-vol.yaml
```diff
@@ -0,0 +1,24 @@
...
+  # The Sage project (Infrastructure, amp-ad, ntap, dream, etc..)
+  Project: "amp-ad"
+  # The resource owner
+  OwnerEmail: "william.poehlman@sagebase.org"
+
+  AvailabilityZone: "us-east-1a"
+  VolumeType: "io1"
+  Size: "1280"
+  Iops: "64000"
+  Ec2InstanceId: "i-0960d784a9185f87e"
```
zaro0508: this instance id is for `wpoehlm-parallel-ec2` instance.  it will need to be updated since the instance was removed by PR https://github.com/Sage-Bionetworks/scicomp-provisioner/pull/131

wpoehlm: When I look at the wpoehlm-iops-ec2 instance (the one we want to attach this volume to), the instance ID is the same as here (i-0960d784a9185f87e)

zaro0508: oops.  you are correct.  I got confused.  no need to update

wpoehlm:
@zaro0508 I'm ready to merge this when you are.

PR #125, opened 2019-04-16
kelsey-analysis

kelshmo:
provision a larger instance for computationally intensive analyses


config/prod/kelsey-analysis.yaml
```diff
@@ -0,0 +1,21 @@
+template_path: managed-ec2-v5.yaml
```
zaro0508: there is a newer version.  https://github.com/Sage-Bionetworks/scicomp-provisioner/pull/3/files

please use the new version of this template.

PR #130, opened 2019-04-17
Wpoehlm ec2

wpoehlm:


zaro0508:
would you mind providing a more descriptive commit message?  Also this will delete the current `wpoehlm-parallel-ec2` ec2 instance and provision a new `wpoehlm-iops-ec2` instance.  Is this what you want?

wpoehlm:
@zaro0508 Will do.  Looks like I might have properly synced with upstream before committing, although I thought it would be alright since that file is the same.  Let me redo the PR

PR #136, opened 2019-04-18
Kelsey windows instance w new deployment

kelshmo:


kelshmo:
I definitely want to keep my `kelsey-analysis` instance and then open a separate windows instance. I thought to request a new instance, I need to use a unique `stack_name`?

zaro0508:
yes, you are using a unique stack name that's not the problem.  What you are doing in this PR is editing an existing file, actually renaming it (config/prod/kelsey-analysis.yaml → config/prod/kelsey-windows.yaml).  What you want to do is to simply add a new `kelsey-windows.yaml` file with the windows configs.

PR #137, opened 2019-04-18
new request clean history

kelshmo:


config/prod/kelsey-windows.yaml
```diff
@@ -0,0 +1,46 @@
+# Provision a Windows EC2 instance
+template_path: remote/managed-ec2-win.yaml
+stack_name: kelsey-windows
+parameters:
+  # The Sage deparment (Platform, CompOnc, SysBio, Governance, etc..)
+  Department: "SysBio"
+  # The Sage project (Infrastructure, amp-ad, ntap, dream, etc..)
+  Project: "Infrastructure"
```
zaro0508: Shouldn’t  this be something specific to sysbio?

kelshmo: It will be used across consortia so I left it vague...

PR #139, opened 2019-04-26
wpoehlm no longer needs large EBS volume with provisioned IOPS

wpoehlm:
volume is unmounted from instance


wpoehlm:
> 
> 
> Is this a generic problem or just for this particular case? Was hoping to use this method for Sara's instance that keeps failing...

I needed this particular volume to get fast data transfer (1 GB/s) to my instance.  The provisioned IOPS are expensive, but this method should work fine for other EBS types.  Wasn't much of a hassle for me, just needed to mount the volume once it was provisioned

wpoehlm:
Found a chunk of files that I thought we would have checksums for already, but we don't.  So, I'm going to use this volume over the weekend after all.

PR #154, opened 2019-05-10
Provision RStudio environment to use for miniDREAM course

milen-sage:
Provision RStudio environment to use for

1) final preparations for the 2019 miniDREAM course
2) production environment for the miniDREAM course

Instance will run RStudio along with its public web interface allowing access to an R environment
for about 20 users on the same machine.

Users will perform calculations on datasets, loading potentially dozens of GBs of data in memory, concurrently. No PHI data will be stored here.

Requested instance: r4.16xlarge

### DIFF
--- a/config/prod/minidream-R-env-pub-prod-instance-ec2.yaml
+++ b/config/prod/minidream-R-env-pub-prod-instance-ec2.yaml
@@ -1,24 +1,46 @@
 # Provision an EC2 instance
-template_path: managed-ec2-v7.yaml
+template_path: remote/managed-ec2.yaml
 stack_name: minidream-R-env-pub-prod-instance-ec2 
 parameters:
-  # The Sage deparment for this resource
+  # The Sage deparment (Platform, CompOnc, SysBio, Governance, etc..)
   Department: "Computational Oncology"
   # The Sage project this resource will be used for
   Project: "CSBC"
-  # The resource owner
   OwnerEmail: "milen.nikolov@sagebase.org"
-  AMIId: "ami-084302eb33c52749d"
-  # EC2 instance type (available types https://aws.amazon.com/ec2/instance-types/)
+
+  # Default account settings, leave alone unless you plan to override
+  VpcName: "{{stack_group_config.default_vpc}}"
+  KeyName: "{{stack_group_config.default_key_pair}}"
+  AMIId: "{{stack_group_config.default_ami}}"
+
+  # Settings have default values but can be overriden. Set the below
+  # parameters *only* if you want to override the defaults.
+
+  # (Optional) EC2 instance type, default is "t2.nano" (other available types https://aws.amazon.com/ec2/instance-types/)
   InstanceType: "r4.16xlarge"
+  # (Optional) EC2 boot volume size in GB, default is 8GB, Max is 1000GB
+  VolumeSize: "500"
+  # (Optional) Base instance on AMI (default: ami-082278746f893d99c) (supported distros: AWS linux)
+  # AMIId: "ami-0de53d8956e8dcf80"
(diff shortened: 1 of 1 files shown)


milen-sage:
Should I use managed-ec2-v7.yaml or managed-ec2-v6.yaml? Just noticed I don't see r4.16xlarge under allowed values in managed-ec2-v6.yaml

zaro0508:
use managed-ec2-v7.yaml

milen-sage:
Done - template changed.

config/prod/minidream-R-env-pub-prod-instance-ec2.yaml
```diff
@@ -0,0 +1,24 @@
+# Provision an EC2 instance
```
zaro0508: no. files in the PRs have changed.  you need to copy from the updated file, https://github.com/Sage-Bionetworks/scicomp-provisioner/pull/3/files

milen-sage: Thanks @zaro0508, I think I used the updated template now.

config/prod/minidream-R-env-pub-prod-instance-ec2.yaml
```diff
@@ -0,0 +1,46 @@
...
+  # parameters *only* if you want to override the defaults.
+
+  # (Optional) EC2 instance type, default is "t2.nano" (other available types https://aws.amazon.com/ec2/instance-types/)
+  InstanceType: "r4.16xlarge"
+  # (Optional) EC2 boot volume size in GB, default is 8GB, Max is 1000GB
+  VolumeSize: "500"
+  # (Optional) Base instance on AMI (default: ami-082278746f893d99c) (supported distros: AWS linux)
+  # AMIId: "ami-0de53d8956e8dcf80"
+  # (Optional) Run EC2 in a subnet (default: PrivateSubnet) (valid values: PrivateSubnet or PublicSubnet)
+  VpcSubnet: "PublicSubnet"
```
zaro0508: Due to new security regulation we don't want to allow any public instances on scicomp.  Please comment this out so that it's private.  We will add a load balancer for your app so that users can reach it.  We will also setup a SSL certificate so your app can be reached with `https` instead of `http`

config/prod/minidream-R-env-pub-prod-instance-ec2.yaml
```diff
@@ -0,0 +1,46 @@
...
+  # (Optional) Expose a port to incoming traffic (default is no open ports) (valid range: 1-65535)
+  OpenPort: "8787"
```
zaro0508: please comment.

config/prod/minidream-R-env-pub-prod-instance-ec2.yaml
milen-sage: Hmm, I thought we discussed that with @ahayden and @xschildw yesterday. Initially (yesterday) that was the reason to go with sandbox, if I understand correctly. Today @ahayden said that scicomp is segmented enough, and we can proceed with having public ip and a web server running there (or maybe I'm misunderstanding)? The public ip is crucial for this project even before the class start, so that external people can access the environment. If scicomp doesn't allow public ip then it's a deal breaker, and I'd have to provision an amazon instance differently. If that's the case, do we have a list of the most relevant/significant issues of running the instance on sandbox (@zaro0508 mentioned accidental deletion, which we definitely do not want; what are other issues?). If the issues are too prohibitive we would probably have to start provision ad hoc outside scicomp and sandbox....

zaro0508: Don't worry, your users will be able to access your application, instead of serving your app from the machine directly we will serve your app up from a load balancer which adds an additional level of security.   Making the instance private is to prevent an attacker from being able to directly access the machine itself.

PR #155, opened 2019-05-14
ALB for minidream.synapse.org

zaro0508:
Setup an application load balancer for milen's minidream app.


zaro0508:
Good idea.  Maybe you can propose a change to the [template ](https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/managed-alb-https.yaml) as a PR and then we can discuss on that PR.

PR #158, opened 2019-05-16
request t3.medium EC2 for wpoehlm

wpoehlm:


zaro0508:
Added support for a1 instance in PR https://github.com/Sage-Bionetworks/aws-infra/pull/130
We'll need to merge that before accepting this PR.  Hold tight.

wpoehlm:
> 
> 
> Added support for a1 instance in PR [Sage-Bionetworks/aws-infra#130](https://github.com/Sage-Bionetworks/aws-infra/pull/130)
> We'll need to merge that before accepting this PR. Hold tight.

No rush, thanks!

zaro0508:
a1 instances only works with Arm based AMIs.  I was hoping that it would just work with our deployment system however that doesn't seem to be the case.  It will require some (maybe lots?) debugging to get Arm to work.  For now, would it be possible to using one of the t3 instances instead?

wpoehlm:
> 
> 
> a1 instances only works with Arm based AMIs. I was hoping that it would just work with our deployment system however that doesn't seem to be the case. It will require some (maybe lots?) debugging to get Arm to work. For now, would it be possible to use one of the t3 instances instead?

Changed to t3.medium

PR #177, opened 2019-06-07
add windows instance for genie migration

xindiguo:


xindiguo:
> both requests look good however we prefer one request per PR. could you please split these into 2 separate PRs?

Changed. Only the windows instance is needed at the moment.,

Issue #161, opened 2019-05-23
Provide code/instructions to validate CF script locally


jaeddy:
It'd be nice to include the relevant scripts (with some usage instructions) in this repo or somewhere easily accessible such that users could check their YAML file and debug any problems _before_ creating a PR.

It looks like some of the scripts are
```yaml
jobs:
  include:
    - stage: validate
      script:
        - yamllint ./config ./templates
        - ./lint_stack.sh -l ./config/prod
        - cfn-lint ./templates/**/*
```

It looks like some of the scripts (or at least `lint_stack.sh` are in [this repo](https://github.com/Sage-Bionetworks/infra-utils). Some of the inputs might need to be reconfigured as URLs instead of relative paths. Not sure if there are other dependencies that would be needed...

PR #186, opened 2019-06-21
fix issue #161: provide info on validating templates

zaro0508:


zaro0508:
this is messed up. creating a new PR

PR #189, opened 2019-06-28
Biaffect

philerooski:


config/prod/phil-biaffect.yaml
```diff
@@ -0,0 +1,47 @@
+# Provision an EC2 instance
+template_path: remote/managed-ec2-v10.j2
+stack_name: iwanna-instance
```
thomasyu888: Update stack name here to yaml filename without the extension

config/prod/phil-biaffect.yaml
```diff
@@ -0,0 +1,47 @@
...
+  KeyName: "scicomp"
+  AMIId: "ami-0de53d8956e8dcf80"      # Amazon Linux 2 AMI 2.0.20190313
+
+  # Settings have default values but can be overriden. Set the below
+  # parameters *only* if you want to override the defaults.
+
+  # (Optional) EC2 instance type, default is "t2.nano" (other available types https://aws.amazon.com/ec2/pricing/on-demand/)
+  # InstanceType: "t2.medium"
+  # (Optional) EC2 boot volume size in GB, default is 8GB, Max is 1000GB
+  # VolumeSize: "50"
```
thomasyu888: Are you ok with small disk size and type?

philerooski: yeah, we'll see how well it works. I'm just downloading and storing a few dozen table rows at a time.

PR #192, opened 2019-07-01
Aws ec2

jgockley62:


config/prod/JKG-Scicomp-Midi.yaml
```diff
@@ -0,0 +1,47 @@
+# Provision an EC2 instance
+template_path: remote/managed-ec2-v10.j2
+stack_name: JKG-Scicomp-Midi2
```
zaro0508: could you please make this match the filename or vise versa?

jgockley62: Should be good to go. Also fixed yaml file extension

PR #199, opened 2019-07-05
Small EC2 instance for udall superusers

philerooski:


thomasyu888:
What is udall superusers?  Are more than one people using this instance?

philerooski:
No, Udall superusers is the name of the mPower 2 substudy. This instance will only be accessed by myself and will be used for adding users to the study and performing some other recurring actions.

PR #202, opened 2019-07-10
Change project tag for WGS bucket

wpoehlm:
We would like to keep track of billing for this project separately, so I have changed the project to "amp-ad-wgs"


wpoehlm:
> 
> 
> I assume changing project tag won't affect the instance.

That is my assumption as well, correct @zaro0508 ?

zaro0508:
no it won't.  although this is a bucket not an instance.

PR #203, opened 2019-07-11
Aws ec2

jgockley62:


jgockley62:
Only adding one more instance, JKGA is already requisitioned and active. This PR is for JKGB.

PR #214, opened 2019-07-20
Revert "adding windows test instance"

thomasyu888:
Reverts Sage-Bionetworks/scicomp-provisioner#213


thomasyu888:
Wrong AMI? @ahayden

zaro0508:
Looks like AWS pulled the rug from under us.  That AMI definitely used to exist because at one point this did provision a win instance: https://github.com/Sage-Bionetworks/sandbox-provisioner/blob/master/config/prod/kd-win-ec2.yaml#L15

I've checked and now it's gone[1].  The latest win AMI is now `ami-04ca2d0801450d495`[2].  Will need to update the example template with that new AMI.

[1] https://pasteboard.co/IoTpZEM.png
[2] https://pasteboard.co/IoTqpE8X.png

PR #128, opened 2019-04-16
Example PR: Auto provision an EC2 Windows instance

zaro0508:
This is an example PR (do not merge!). It is a template to show how to get
the CI system to automatically provision a Windows 2019 Server EC2 instance.

Instance creation workflow:

* Scientist create a PR similar to this one. Create a new file
* with a unique stack_name and updated parameters.
* Have someone review and approve the PR.
* Merge the PR
* The CI system will auto provision an EC2 instance
* An email will be sent to the owner email address with the new bucket info.

The newly created instance will contain the following tags:
```
key: Department
value: Platform

key: OwnerEmail
value: joe.smith@sagebase.org

key: Project
value: Infrastructure

key: aws:cloudformation:logical-id
value: Ec2Instance

key: aws:cloudformation:stack-id
value: arn:aws:cloudformation:us-east-1::stack/iwanna-instance/559e55a0-bc47-11e8-995a-500c28b4b099

key: aws:cloudformation:stack-name
value: iwanna-instance
```

To remove the instance:

* Instance owner must create a new PR that deletes the files that was added by
the PR that was used to add the instance.
* Have an an AWS account Admin review your change.
* Approve and merge.

PR #218, opened 2019-07-25
adding testhost win

ahayden:
I stopped doing this last week, but am back to it... the AMI ID was wrong last time, but should be good now.

### DIFF
--- a/config/prod/ahayden-testhost-win.yaml
+++ b/config/prod/ahayden-testhost-win.yaml
@@ -1,0 +1,50 @@
+# Provision a Windows EC2 instance
+template_path: remote/managed-ec2-win-v4.j2
+stack_name: ahayden-testhost-win
+parameters:
+  # The Sage deparment (Platform, CompOnc, SysBio, Governance, etc..)
+  Department: "Platform"
+  # The Sage project (Infrastructure, amp-ad, ntap, dream, etc..)
+  Project: "Infrastructure"
+  # The resource owner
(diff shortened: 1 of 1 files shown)


config/prod/ahayden-testhost-win.yaml
```diff
@@ -0,0 +1,50 @@
...
+  Department: "Platform"
+  # The Sage project (Infrastructure, amp-ad, ntap, dream, etc..)
+  Project: "Infrastructure"
+  # The resource owner
+  OwnerEmail: "aaron.hayden@sagebase.org"
+
+  # Default account settings, leave alone unless you plan to override
+  VpcName: "computevpc"
+  KeyName: "scicomp"
+  AMIId: "ami-0388ce15d3f7066cb"    # Windows_Server-2019-English-Full-2019.03.13
```
zaro0508: this is not the same AMI as in the example template.  https://github.com/Sage-Bionetworks/scicomp-provisioner/pull/128/files#diff-f5e07fc328d4bdc4c00d094873166699R15

if you want this AMI could you please add a comment with info about it?

ahayden: It's the default Win AMI in config.yaml?... Remember the switcharoo with AMIs? Could it be the example template does not have the AMI ID that AWS lists after the switcharoo?

zaro0508: i have updated the example template.  https://imgur.com/a/ANVefVU

PR #219, opened 2019-07-28
Add testhost win

ahayden:
Let's try again with the JC agent dependency change


ahayden:
might have merged too quick...

PR #223, opened 2019-08-20
Aws ec2 RInstance

jgockley62:


config/prod/JKGRinst.yaml
```diff
@@ -0,0 +1,47 @@
+# Provision an EC2 instance
+template_path: remote/managed-ec2-v10.j2
```
zaro0508: this is not the latest template.  please take another look at the example template

jgockley62: Changed to remote/managed-ec2-linux-v1.j2

PR #224, opened 2019-08-21
Alter Instance Type and Storage Size

jgockley62:


jgockley62:
Sry forgot to delete comments hash symbol for instance type and size

config/prod/JKGRinst.yaml
```diff
@@ -23,9 +23,9 @@ parameters:
   # parameters *only* if you want to override the defaults.
 
   # (Optional) EC2 instance type, default is "t2.nano" (other available types https://aws.amazon.com/ec2/pricing/on-demand/)
-  # InstanceType: "r4.4xlarge"
+  InstanceType: "r4.4xlarge"
   # (Optional) EC2 boot volume size in GB, default is 8GB, Max is 1000GB
-  # VolumeSize: "180"
+  VolumeSize: "180"
```
zaro0508: This change will wipe data on your current instance.  Is that ok?

jgockley62: Yeah all good

PR #230, opened 2019-08-29
Cmc s3 new

kelshmo:
Updated with ARN

### DIFF
--- a/config/prod/CMC-large-s3.yaml
+++ b/config/prod/CMC-large-s3.yaml
@@ -1,0 +1,41 @@
+# Provision a Synapse External Bucket (http://docs.synapse.org/articles/custom_storage_location.html)
+template_path: remote/s3-bucket.yaml
+stack_name: amp-mayo-sinai
+parameters:
+  # The Sage deparment (Platform, CompOnc, SysBio, Governance, etc..)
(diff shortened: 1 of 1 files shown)


config/prod/CMC-large-s3.yaml
```diff
@@ -0,0 +1,39 @@
...
+  # You may omit them if you do not need to override the defaults.
+
+  # (Optional) true for read-write bucket, false (default) for read-only bucket
+  AllowWriteBucket: 'true'
+  # (Optional) Synapse username (default: ""), required if AllowWriteBucket=true
+  SynapseUserName: 'kelsey'
+  # (Optional) Allow accounts, groups, and users to access bucket (default is no access).
+  GrantAccess:
+    - 'arn:aws:iam::325565585839:root'   # Required ARN for a synapse bucket
+    - 'arn:aws:iam::945062467230:user/kelsey.montgomery@sagebase.org'
```
zaro0508: is this your personal AWS account because that's not a Sage AWS account ID.  i believe this is OK just as long as that ARN does indeed exists.

kelshmo: I updated the ARN to the Sage AWS account ID.

PR #241, opened 2019-09-18
Kelsey iops

kelshmo:
I increased the volume size and changed the instance type to be better suited for this use case (and save money!)


kelshmo:
per @thomasyu888's comment via slack, it is ok to wipe the instance.

PR #242, opened 2019-09-27
EC2 for pku analysis daily job

philerooski:


config/prod/phil-pku-analysis.yaml
```diff
@@ -0,0 +1,47 @@
+# Provision an EC2 instance
+template_path: remote/managed-ec2-v10.j2
```
thomasyu888: Wrong version

config/prod/phil-pku-analysis.yaml
```diff
@@ -0,0 +1,47 @@
...
+stack_name: pku-analysis
```
thomasyu888: Stack name should make file name. `stackname`.yaml

philerooski: Is this information written down anywhere?

thomasyu888: No, its just "easier" to keep track of the stack-names if they match the filename.  They don't *have* to match though.

thomasyu888: Actually it is written down on this page: https://sagebionetworks.jira.com/wiki/spaces/IT/pages/708083719/Sage+AWS+Provisioner

> With GitHub access you can, you can deploy a new instance by creating a new yaml file. Copy the contents of an example like this (in raw) into a new file, such as: ~/src/sandbox-provisioner/config/prod/new-test-instance-for-$USER.yaml. Make sure to use the newest available version of the template that will define your instance. Also note that this file is not in the master branch of the repository, but instead resides in a pull request. The template indicates optional files that may be changed (e.g., InstanceType and VolumeSize). Additionally, you may want to change the AMI (using the AMIId variable). **Finally, you want to change the stack_name to be the same as the prefix of the yaml file** and the Department, Project, and OwnerEmail parameters.

philerooski: 🤦‍♂

config/prod/phil-pku-analysis.yaml
```diff
@@ -0,0 +1,48 @@
+# Provision an EC2 instance
+template_path: remote/managed-ec2-linux-v1.j2
+stack_name: iwanna-instance
```
thomasyu888: Update stack name to `phil-pki-analysis`

philerooski:
@thomasyu888 We need to advocate harder for your web UI implementation of this mess 😥

philerooski:
Too many opportunities for human error.

PR #258, opened 2019-10-19
Added workstation for Vanessa

ahayden:


zaro0508:
Ugh! why are we provisioning on behalf of scientists?

ahayden:
Because IT-714

PR #262, opened 2019-11-16
HTAN data model visualization website instance

milen-sage:
Request an instance optimized to run a web site with an elastic search backend:
two open ports 8000 and 9200.

### DIFF
--- a/config/prod/htan-data-model-web-ec2.yaml
+++ b/config/prod/htan-data-model-web-ec2.yaml
@@ -1,6 +1,10 @@
 # Provision an EC2 instance
-template_path: remote/managed-ec2-v10.j2
+template_path: remote/managed-ec2-linux-v1.j2
 stack_name: htan-data-model-web-ec2
+sceptre_user_data:
+  Distro: "aws"
+  # (Optional) Expose ports to incoming traffic (default is no open ports) (valid range: 1-65535)
+  OpenPorts: ["8000", "9200"]
 parameters:
   # The Sage deparment (Platform, CompOnc, SysBio, Governance, etc..)
   Department: "CompOnc"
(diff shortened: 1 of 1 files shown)


milen-sage:
> Use new template.

I think I used the correct one now?

config/prod/htan-data-model-web-ec2.yaml
```diff
@@ -0,0 +1,50 @@
+# Provision an EC2 instance
+template_path: remote/managed-ec2-linux-v1.j2
+stack_name: htan-data-model-web-ec2
+sceptre_user_data:
+  Distro: "aws"
+  # (Optional) Expose ports to incoming traffic (default is no open ports) (valid range: 1-65535)
+  OpenPorts: ["8000", "9200"]
```
zaro0508: this is ok.  was wondering if this website should only be accessible by sage employees or should it be accessible by people outside of sage as well?

config/prod/htan-data-model-web-ec2.yaml
```diff
@@ -0,0 +1,50 @@
...
+  # parameters *only* if you want to override the defaults.
+
+  # (Optional) EC2 instance type, default is "t2.nano" (other available types https://aws.amazon.com/ec2/instance-types/)
+  InstanceType: "t2.medium"
+  # (Optional) EC2 boot volume size in GB, default is 8GB, Max is 1000GB
+  VolumeSize: "24"
+  # (Optional) Base instance on AMI (default: ami-082278746f893d99c) (supported distros: AWS linux)
+  # AMIId: "ami-0de53d8956e8dcf80"
+  # (Optional) Run EC2 in a subnet (default: PrivateSubnet) (valid values: PrivateSubnet or PublicSubnet)
+  VpcSubnet: "PublicSubnet"
```
zaro0508: if this is only for sage employees then we prefer "PrivateSubnet"

config/prod/htan-data-model-web-ec2.yaml
milen-sage: The site should be accessible outside Sage as well.
It'd be great to assign a domain name as well eventually, we'll have to decide on the name first w/in HTAN.

PR #263, opened 2019-11-19
Test template changes

ahayden:
Provision a few instance types to test management and the templates


config/prod/ahayden-update-test-linv1.yaml
```diff
@@ -0,0 +1,48 @@
+# Provision an EC2 instance
+template_path: remote/managed-ec2-linux-v1.j2
```
zaro0508, line 2: this is an old template.  please use latest example, https://github.com/Sage-Bionetworks/scicomp-provisioner/pull/3

PR #14, opened 2018-11-12
Example PR: Auto provision a S3 bucket

zaro0508:
This is an example PR (do not merge!). It is a template to show
how to get the CI system to automatically provision a bucket.

Bucket creation workflow:

* Scientist create a PR similar to this one. Create a new file
  with a unique stack_name and updated parameters.
* Have someone review and approve the PR.
* Merge the PR
* The CI system will auto provision the bucket
* An email will be sent to the bucket owner email address with
  the new bucket info.

The newly created bucket will contain the following tags:

```
key: aws:cloudformation:stack-id
value: arn:aws:cloudformation:us-east-1:<AWS account ID>:stack/GatesKI-TestStudy/375ced80-78b9-11e8-bc55-500c286e1a36

key: aws:cloudformation:stack-name
value: GatesKI-TestStudy

key: aws:cloudformation:logical-id
value: SynapseExternalBucket

key: OwnerEmail
value: joe.smith@sagebase.org
```

To remove the bucket:

* Bucket owner creates a new PR that deletes the files that was added by
  the PR that was used to add the bucket.
* Have an an AWS account Admin review your change.
* Approve and merge.
* Since cloudformation does not allow deleting a buckets with data in it
  the bucket owner will need to contact an admin to delete the bucket
  on their behalf.

PR #273, opened 2020-01-28
Create a custom synapse bucket for HTAN project

zaro0508:
Create a synapse bucket for HTAN collaboration.  The idea is that
a bunch of HTAN collaborator will read and write to this bucket.
Folders will be created for each colloborator and permissions will
be configure for each collobartor to write to a folder.

### DIFF
--- a/templates/htan-synapse-s3-storage.yaml
+++ b/templates/htan-synapse-s3-storage.yaml
@@ -1,0 +1,95 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Transform: S3Objects
+Description: 'Synapse S3 Custom Storage (https://docs.synapse.org/articles/custom_storage_location.html)'
+Metadata:
+  AWS::CloudFormation::Interface:
+    ParameterGroups:
+      - Label:
+          default: S3 Bucket Policy Configuration
+        Parameters:
+          - BucketName
+          - SynapseUserName
+          - S3UserARNs
+Parameters:
+  SynapseUserName:
+    Type: String
+    Description: Synapse user for bucket
+    Default: ""
+  S3UserARNs:
+    Type: CommaDelimitedList
+    Default: ""
+    Description: User ARNs allowed to access S3 Bucket
+    ConstraintDescription: List of ARNs separated by commas (i.e. arn:aws:iam::011223344556:user/jsmith,arn:aws:iam::544332211006:user/rjones)
+  BucketName:
+    Type: String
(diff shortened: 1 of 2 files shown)


templates/htan-synapse-s3-storage.yaml
```diff
@@ -0,0 +1,95 @@
...
+        config:
+          ignore_checks:
+            - E3001
+    Properties:
+      Target:
+        Bucket: !Ref S3Bucket
+        Key: owner.txt
+        ContentType: text
+        ACL: authenticated-read
+      Body: !Ref SynapseUserName
```
zaro0508: Wondering if synapse supports multiple synapse users accessing one bucket?  The documentation seems to indicate that it doe not support this use case.

xdoan: can a team be used instead of the username? or some type of service account? If we're creating custom storage locations (due to file size and synapseclient upload bottlenecks) we wouldn't suggest they upload through synapse, though try they might wonder why they can't.

xschildw: There's an issue opened re: multiple users in owner.txt: https://sagebionetworks.jira.com/browse/PLFM-3590

xdoan: Thank you! Just commented on the jira.

templates/htan-synapse-s3-storage.yaml
```diff
@@ -0,0 +1,95 @@
...
+  AWS::CloudFormation::Interface:
+    ParameterGroups:
+      - Label:
+          default: S3 Bucket Policy Configuration
+        Parameters:
+          - BucketName
+          - SynapseUserName
+          - S3UserARNs
+Parameters:
+  SynapseUserName:
```
xschildw: We're supporting userid/teamid and encourage migrating to it. Could we renam param and change description to Synapse userId or teamId. Technically we also support multiple but I think that would complicate more than it's worth.

zaro0508: here's the issue @xschildw https://sagebionetworks.jira.com/browse/IT-831

would you like to work on it?

zaro0508:
The PR provisioner supports multiple synapse users and teams now @xdoan.  To get that functionality use the new (v2) [example template](https://github.com/Sage-Bionetworks/scicomp-provisioner/pull/14/files) and set the SynapseIDs configuration.

PR #277, opened 2020-02-11
Adds danWebster instance for scicomp

DanWebster:
Using for general compute with potential prod use

### DIFF
--- a/config/prod/dwebster-scicomp-general.yaml
+++ b/config/prod/dwebster-scicomp-general.yaml
@@ -1,0 +1,48 @@
+# Provision an EC2 instance
+template_path: remote/managed-ec2-linux-v1.j2
+stack_name: dwebster-scicomp-general
+sceptre_user_data:
+  Distro: "aws"     # (valid values: aws, ubuntu) AMIId must match distro
+  # (Optional) Expose ports to incoming traffic (default is no open ports) (valid range: 1-65535)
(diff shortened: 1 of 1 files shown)


config/prod/dwebster-scicomp-general.yaml
```diff
@@ -0,0 +1,48 @@
+# Provision an EC2 instance
+template_path: remote/managed-ec2-linux-v1.j2
+stack_name: dwebster-scicomp-general
+sceptre_user_data:
+  Distro: "aws"     # (valid values: aws, ubuntu) AMIId must match distro
+  # (Optional) Expose ports to incoming traffic (default is no open ports) (valid range: 1-65535)
+  OpenPorts: ["80","443"]
```
zaro0508: By default all instances are private, meaning you can only access it after logging into the VPN.  This setting is only needed if you want a public instance, your instance to be accessible from anywhere on the internet.  do you need that?  if you do then you will also need to set `VpcSubnet`  to PublicSubnet.  if you don't then keep this as a comment.

ahayden: We want a private instance with those port open, so we'll revise the PR to not define the open ports.

zaro0508: all ports are open for private instances

PR #279, opened 2020-02-12
Add emory pipeline lambda stacks

ecmonsen:
Copied templates from sandbox-provisioner, and updated template path and stack parameters for production.

### DIFF
--- a/templates/ecmonsen-emorypipeline-ci.yaml
+++ b/templates/ecmonsen-emorypipeline-ci.yaml
@@ -1,0 +1,140 @@
+Description: Setup service account and roles for Emory pipeline lambda
+AWSTemplateFormatVersion: 2010-09-09
+
+Parameters:
+  Department:
+    Description: 'The department for this resource'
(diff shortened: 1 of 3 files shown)


templates/ecmonsen-emorypipeline-ci.yaml
```diff
@@ -0,0 +1,140 @@
+Description: Setup service account and roles for Emory pipeline lambda
```
zaro0508: this same file is duplicated in both sandbox-provisioner and scicomp-provisioner.  would it make sense to move this to `Sage-Bionetworks/ecmonsen-emorypipeline-lambda-code` repro?  This would allow you to  deploy it to your public bucket and then shared to sandbox-provisioner and scicomp-provisioner?

ecmonsen:
In theory, yes, that makes so much sense. In practice, I'm pretty eager to
get this deployed without any more big changes. I'll create an issue in the
code repo to fix this later.

PR #302, opened 2020-02-25
decrease instance size for writeup phase

vpchung:


config/prod/verena-allen-chal.yaml
```diff
@@ -16,16 +16,16 @@ parameters:
...
 
   # Settings have default values but can be overriden. Set the below
   # parameters *only* if you want to override the defaults.
 
   # (Optional) EC2 instance type, default is "t2.nano" (other available types https://aws.amazon.com/ec2/pricing/on-demand/)
-  InstanceType: "t3.2xlarge"
+  InstanceType: "t3.small"
   # (Optional) EC2 boot volume size in GB, default is 8GB, Max is 1000GB
-  VolumeSize: "200"
+  # VolumeSize: "200"
```
thomasyu888: Leave this at 200 or it will wipe your instance.

vpchung: okay, thanks!

PR #303, opened 2020-02-26
Managed rds postgres

ahayden:
Please do not merge, this is for reference to another PR.


config/prod/test-ahayden-rds-postgres.yaml
```diff
@@ -0,0 +1,43 @@
+# Provision an EC2 instance
+template_path: remote/managed-rds-postgres-v1.j2
+stack_name: test-ahayden-rds-postgres
+sceptre_user_data:
+  # (Optional) Expose database ports to this whitelist of CIDR blocks
+   OpenIngress: ["207.109.76.122"]
```
xschildw: Should it not be gated by the VPN?

ahayden: Right? Nope, this is layer three whitelisted to an app on the web.

config/prod/test-ahayden-rds-postgres.yaml
```diff
@@ -0,0 +1,43 @@
...
+  # Set unique database user and password values in ssm with a path specific to the project
+  DBMasterUsername: !ssm /iAtlas/postgresql/dev/test-ahayden-rds-postgres-username
+  DBMasterUserPassword: !ssm /iAtlas/postgresql/dev/test-ahayden-rds-postgres-password
+  #
+  # Set SSL/TLS CSR values
+  DnsDomainName: "sagebionetworks.org"
+  DnsSubDomainName: "iatlasdb"
+
+  # Default account settings, leave alone unless you plan to override
+  VpcName: "snowflakevpc"
```
ahayden: @zaro0508 this is where I use the VPC name

tthyer: @ahayden @zaro0508 is it a good idea for other things to share `snowflakevpc` with the toil cluster? Considering the extensive permissions that the cluster leader has

ahayden: I think you're right... computevpc instead?

PR #304, opened 2020-02-27
Request for EC2 with RStudio for amp-ad data curation tasks

Aryllen:
Title/commit message says it all.


config/prod/nkauer-data-curation.yaml
```diff
@@ -0,0 +1,48 @@
+# Provision an EC2 instance
+template_path: remote/managed-ec2-linux-v1.j2
+stack_name: nkauer-r-studio
```
zaro0508: please name it the same as the file name or vice versa

Aryllen: Whoops. Renamed.

PR #309, opened 2020-03-13
Remove instance

thomasyu888:


thomasyu888:
Yup. Going to switch everything to service catalog

PR #311, opened 2020-03-25
Provision RDS for iAtlas project

ahayden:
This template requires manual set up before merging:
(1) Complete whitelist definition and update branch
(2) Create ACM cert and overwrite placeholder for ARN value and update branch
(3) Add secret for DB user to SSM

Depends: https://github.com/Sage-Bionetworks/aws-infra/pull/226


ahayden:
This DB needs a public address as the request of the designers... The whitelist is a list object that jinja iterates through to make security groups.

ahayden:
Closing because we found a better solution.

PR #323, opened 2020-04-24
Increase GENIE nexus resources

thomasyu888:
I know this will delete everything in the instance.  I am ok with it.


config/prod/tom-genome-nexus.yaml
```diff
@@ -23,9 +23,9 @@ parameters:
   # parameters *only* if you want to override the defaults.
 
   # (Optional) EC2 instance type, default is "t3.micro" (other available types https://aws.amazon.com/ec2/pricing/on-demand/)
-  # InstanceType: "t3.medium"
+  InstanceType: "t3.medium"
   # (Optional) EC2 boot volume size in GB, default is 8GB, Max is 1000GB
-  # VolumeSize: "50"
+  VolumeSize: "200"
```
thomasyu888, line 28:
```suggestion
  VolumeSize: "100"
```

PR #326, opened 2020-04-27
change owner to larsson omberg

kdaily:
Note that the `SynapseUserName` is not changed and should be adjusted later once it's known how to manage that.

### DIFF
--- a/config/prod/mep-lincs-s3.yaml
+++ b/config/prod/mep-lincs-s3.yaml
@@ -14,7 +14,7 @@ parameters:
   OwnerEmail: "larsson.omberg@sagebase.org"
   # (Optional) true for read-write bucket, false (default) for read-only bucket
   AllowWriteBucket: 'true'
-  SynapseUserName: 'kdaily'
+  SynapseUserName: 'larssono'
   GrantAccess:
     - 'arn:aws:iam::325565585839:root'   # Required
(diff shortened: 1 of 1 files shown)


zaro0508:
@kdaily please update this PR with the change to SynapseUserName that you want.   We update the SynapseUserName automatically in newer versions of the template but since this uses an older version we need to update it manually. 

Whichever sageit admin merges this PR should also update this bucket's owner.txt file with the new  SynapseUserName setting.

config/prod/mep-lincs-s3.yaml
```diff
@@ -11,7 +11,7 @@ parameters:
   # The Sage project (Infrastructure, amp-ad, ntap, dream, etc..)
   Project: "MEP-LINCS"
   # The resource owner
-  OwnerEmail: "kenneth.daily@sagebase.org"
+  OwnerEmail: "larsson.omberg@sagebase.org"
   # (Optional) true for read-write bucket, false (default) for read-only bucket
   AllowWriteBucket: 'true'
   SynapseUserName: 'kdaily'
```
thomasyu888, line 17:
```suggestion
  SynapseUserName: 'larssono'
```

thomasyu888:
@zaro0508 I patched

PR #343, opened 2020-06-23
add phil-cough-data stack

philerooski:
Create S3 bucket for cough data data transfer.

### DIFF
--- a/config/prod/phil-cough-data.yaml
+++ b/config/prod/phil-cough-data.yaml
@@ -18,7 +18,7 @@ parameters:
   # (Optional) Synapse username (default ""), required if AllowWriteBucket=true
   SynapseUserName: 'phil'
   # (Optional) Allow accounts, groups, and users to access bucket. (default is no access).
-   GrantAccess:
+  GrantAccess:
      - 'arn:aws:iam::325565585839:root'   # Required ARN for a synapse bucket
      - 'arn:aws:s3:::hyfe-sage' # Joe Brew ARN
   # (Optional) true (default) to encrypt bucket, false for no encryption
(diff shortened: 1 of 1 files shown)


philerooski:
It's a lot of data and I assume it's related to a grant. We've tied specific buckets to specific grants in the past for expense tracking.

config/prod/phil-cough-data.yaml
```diff
@@ -0,0 +1,42 @@
+# Provision a general S3 bucket or a Synapse External Bucket
+# http://docs.synapse.org/articles/custom_storage_location.html
+template_path: remote/s3-bucket.yaml
+stack_name: cough-data
```
zaro0508, line 4: stack_name and filename must match.

config/prod/phil-cough-data.yaml
```diff
@@ -0,0 +1,42 @@
+# Provision a general S3 bucket or a Synapse External Bucket
+# http://docs.synapse.org/articles/custom_storage_location.html
+template_path: remote/s3-bucket.yaml
+stack_name: phil-cough-data
```
zaro0508, line 4: ohh wait, i think you said this bucket will likely live longer than you?  do you really want to have your name on the resources then?  AWS will append this `stack_name` to all the resources it creates.

philerooski: Hahaha please just merge it 😂 This will be my legacy.

philerooski:
@zaro0508 If you're still confused about why I opened the PR here rather than using scipool can you reach out to science LT and talk with them about what SOP is going forward with AWS resources that are tied to grants now that scipool and the service catalog are on the scene? I'm using what I believe to be SOP as of a few months ago but if it's changed in the meanwhile we should propagate that info to the rest of science.

PR #344, opened 2020-06-26
ShinyserverPro instance production

xschildw:
This is the instance to run shinyserver-pro production server.

### DIFF
--- a/config/prod/shinyserver-pro-prod.yaml
+++ b/config/prod/shinyserver-pro-prod.yaml
@@ -1,0 +1,46 @@
+# Provision an EC2 instance
+template_path: remote/managed-ec2-linux-v2.j2
+stack_name: shinyserver-pro-prod
+sceptre_user_data:
+  Distro: "ubuntu"     # (valid values: aws, ubuntu) AMIId must match distro
+  # (Optional) Expose ports to incoming traffic (default is no open ports) (valid range: 1-65535)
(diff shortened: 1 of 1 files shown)


config/prod/shinyserver-pro-prod.yaml
```diff
@@ -0,0 +1,46 @@
...
+  # The Sage project (Infrastructure, amp-ad, ntap, dream, etc..)
+  Project: "Infra"
+  # The resource owner
+  OwnerEmail: "x.schildwachter@sagebase.org"
+
+  # Default account settings, leave alone unless you plan to override
+  VpcName: "computevpc"
+  KeyName: "scicomp"
+  # AMIId: "ami-0ca7208f904dc1874"      # AWS linux 2 AMI 2.0.20200108 x86_64 HVM gp2
+  AMIId: "ami-0491993ca183b00aa"      # packer-shinyserver-pro-master
```
zaro0508, line 20: packer-shinyserver-pro-master should only be used for testing, if you plan to use this for prod then it's better to create a versioned  packer-shinyserver-pro-master and use that AMI instead.

xschildw: OK, so I understand I need to tag 'v1.0.0'. Do I just pull master, git tag -a then push --tags back to master?

zaro0508:
yes, that sounds right

PR #346, opened 2020-06-29
Phil cough data stack update

philerooski:


config/prod/phil-cough-data.yaml
```diff
@@ -0,0 +1,42 @@
+# Provision a general S3 bucket or a Synapse External Bucket
+# http://docs.synapse.org/articles/custom_storage_location.html
+template_path: remote/s3-bucket.yaml
+stack_name: cough-data
```
zaro0508, line 4: this and filename should match

philerooski: Ugh, I wish there was a way to make this apart of the pre-commit check 🤦

PR #350, opened 2020-06-30
add phil ARN to phil-cough-data bucket access

philerooski:


config/prod/phil-cough-data.yaml
```diff
@@ -21,6 +21,7 @@ parameters:
   GrantAccess:
      - 'arn:aws:iam::325565585839:root'   # Required ARN for a synapse bucket
      - 'arn:aws:iam::671670783497:user/joebrew' # Joe Brew ARN
+     - 'arn:aws:iam::563295687221:user/phil.snyder@sagebase.org' # Phil Snyder ARN
```
zaro0508, line 24: The way to access AWS is thru jumpcloud now so the ARN for you should be `arn:aws:sts::563295687221:assumed-role/sandbox-developer/phil.snyder@sagebase.org`

Read more about it here: https://sagebionetworks.jira.com/wiki/spaces/IT/pages/938836322/Service+Catalog+Provisioning#Which-ARN-to-use?

philerooski: I long for the day we finally have a stable provisioning system 😄

PR #351, opened 2020-07-01
[IT-922] Update GrantAccess config for S3 buckets

zaro0508:
Now that we only allow access to AWS thru jumpcloud we need to
update all the GrantAccess configuration in S3 Buckets to the
new assumed role ARN


zaro0508:
Nope, absolutely not @ahayden.  They will need to look at these for examples or be lucky enough to find a few references in our wiki docs.

PR #358, opened 2020-07-28
remove non-functional instance

kelshmo:
I would like to delete this instance from jumpcloud.


zaro0508:
this is replaced by #360

PR #366, opened 2020-08-18
[IT-937] Setup KMS key for HTAN S3 Synapse Sync

zaro0508:
The S3 Synapse Sync project[1] needs to store secrets in the
AWS SSM. This will create a KMS key to allow us to store those
secrets. The only users that will have access to the key are
AWS account admins and CI systems.

[1] https://github.com/Sage-Bionetworks/s3-synapse-sync

### DIFF
--- a/templates/htan-synapse-sync-kms-key.yaml
+++ b/templates/htan-synapse-sync-kms-key.yaml
@@ -1,0 +1,90 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: HTAN SynapseSync KMS Key and IAM policy
+
+Parameters:
+  AdminRoleArn:
+    Type: String
+    Description: ARN of the Administrator Role for a particular account
+
+Resources:
+  KmsDecryptPolicy:
+    Type: AWS::IAM::ManagedPolicy
+    Properties:
+      PolicyDocument:
+        Version: 2012-10-17
+        Statement:
+          - Sid: ReadAccess
+            Action:
+              - 'kms:Describe*'
+              - 'kms:GetKeyPolicy'
+              - 'kms:GetKeyRotationStatus'
+              - 'kms:List*'
+              - 'kms:Verify'
+            Effect: Allow
+            Resource: !GetAtt KmsKey.Arn
+          - Sid: DecryptAccess
+            Action:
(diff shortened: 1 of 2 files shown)


templates/htan-synapse-sync-kms-key.yaml
```diff
@@ -0,0 +1,90 @@
...
+            Action:
+              - 'kms:*'
+            Resource: '*'
+          - Sid: "Allow administration of the key to CFN service role"
+            Effect: "Allow"
+            Principal:
+              AWS:
+                - !ImportValue
+                    'Fn::Sub': '${AWS::Region}-bootstrap-CfServiceRoleArn'
+                - !Ref AdminRoleArn
```
xschildw, line 52: Can't we do an import as well from admin saml role?

zaro0508: I don't understand your question @xschildw.  Which role specifically are you talking about?  Also what would be the purpose?  pay special attention to the parameters in the config/prod/htan-synapse-sync-kms-key.yaml file and what `AdminRoleArn` resolves to.  This config give scicomp jumpcloud admins access to this key.

xschildw: @zaro0508 , you're right I missed the config part. That's the role I was referring to, you're doing the same as import without coupling the templates. Sorry for the confusion.

templates/htan-synapse-sync-kms-key.yaml
```diff
@@ -0,0 +1,90 @@
...
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-KmsKey'
+  KmsKeyAlias:
+    Value: !Ref KmsKeyAlias
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-KmsKeyAlias'
+  KmsKeyArn:
+    Value: !GetAtt KmsKey.Arn
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-KmsKeyArn'
+  KmsDecryptPolicyArn:
```
ahayden, line 87: This will be imported into a role to allow use of the key.

PR #367, opened 2020-08-19
[IT-937] Add cloudformation template for HTAN bucket

zaro0508:
Add a cloudformation template to automate deployment of
HTAN buckets that will work with the s3-synapse-sync lambda[1]

[1] https://github.com/Sage-Bionetworks/s3-synapse-sync

Depends on Sage-Bionetworks/s3-synapse-sync#3

### DIFF
--- a/templates/htan-synapse-bucket.yaml
+++ b/templates/htan-synapse-bucket.yaml
@@ -1,0 +1,157 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Transform: S3Objects
+Description: >-
+  Synapse S3 Custom Storage
+  (https://docs.synapse.org/articles/custom_storage_location.html)
+Metadata:
+  AWS::CloudFormation::Interface:
+    ParameterGroups:
+      - Label:
+          default: S3 Bucket Configuration
+        Parameters:
+          - BucketName
+      - Label:
+          default: S3 Bucket Policy Configuration
+        Parameters:
+          - SynapseIDs
+          - S3UserARNs
+Parameters:
+  S3SynapseSyncFunctionRoleArn:
+    Type: String
+    Description: The S3 Synapse sync lambda function role ARN
+    ConstraintDescription: >-
+      Must be the lambda role ARN (i.e. arn:aws:iam::055273631518:role/s3-syanpase-sync-LambdaRole-JRQLRZDI9TMS)
+  SynapseRootArn:
+    Type: String
+    Description: The Synapse root account ARN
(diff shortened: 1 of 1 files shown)


zaro0508:
please review @clarisse-lau

templates/htan-synapse-bucket.yaml
```diff
@@ -0,0 +1,157 @@
...
+    Properties:
+      Bucket: !Ref S3Bucket
+      PolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          -
+            Sid: "ReadAccess"
+            Effect: "Allow"
+            Principal:
+              AWS: !Ref S3UserARNs
```
clarisse-lau, line 90: Should the Synapse root arn be included in this principal(and that of the "WriteAccess" statement)?

templates/htan-synapse-bucket.yaml
```diff
@@ -0,0 +1,157 @@
...
+            Action:
+              - "s3:ListBucket*"
+              - "s3:GetBucketLocation"
+            Resource: !Split
+              - ","
+              - !Sub
+                - "${list},${SynapseRootArn}"
+                - list: !Join [",",!Ref "S3UserARNs"]
```
clarisse-lau, line 98: wondering if the resource for “ReadAccess” and “WriteAccess” statements should be the S3 bucket?

templates/htan-synapse-bucket.yaml
```diff
@@ -0,0 +1,157 @@
...
+          -
+            Sid: "LambdaAccess"
+            Effect: Allow
+            Principal:
+              AWS: !Ref S3SynapseSyncFunctionRoleArn
```
clarisse-lau, line 116: I believe S3UserARNs will also need these permissions

zaro0508: "WriteAccess" already provides `s3:*Object*` to S3UserARNs so I don't think we need it here.

templates/htan-synapse-bucket.yaml
zaro0508: done

templates/htan-synapse-bucket.yaml
zaro0508: done

templates/htan-synapse-bucket.yaml
clarisse-lau: Right. Looks good to me then

PR #372, opened 2020-08-22
[IT-937] Update deny policy on htan bucket

zaro0508:
The deny policy on htan bucket policy for s3-synapse-sync buckets
causes access denied on changes to owner.txt file.

Our CFN template creates a bucket with an owner.txt file then applies
the bucket policy which includes a deny unless the PUT includes
the S3 canonical identifiers.  The file is added on initial bucket
creation.  However when the s3objects-macro attempts to update owner.txt
file it gets a access denied error because the macro doesn't do a PUT with
the identifiers.  We need to update the policy to exclude our service user.

depends on Sage-Bionetworks/aws-infra#258


templates/htan-synapse-bucket.yaml
```diff
@@ -107,20 +107,21 @@ Resources:
...
-            Principal: "*"
-            Action: s3:PutObject
-            Resource: !Sub "${S3Bucket.Arn}/*"
-            Condition:
-              StringNotEqualsIgnoreCase:
-                s3:x-amz-grant-full-control: !Sub "id=${SynapseCanonicalId},id=${S3CanonicalUserId}"
-              StringNotEquals:
-                s3:x-amz-grant-full-control: !Sub "id=${SynapseCanonicalId},id=${S3CanonicalUserId}"
+#          - Effect: Deny
+#            Principal: "*"
```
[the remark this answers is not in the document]
clarisse-lau: We could also try using the [NotPrincipal](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_notprincipal.html) element to deny access to all principals except a specified role.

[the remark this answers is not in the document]
zaro0508: Another option is to remove owner.txt file management from this template altogether.  Removing the owner.txt file from this template means that only the synapse bucket and policy gets created when deploying an HTAN bucket with this template.  Once created there would need to be an additional manual step to upload a valid owner.txt file for the bucket.

templates/htan-synapse-bucket.yaml
```diff
@@ -107,20 +107,21 @@ Resources:
...
+#            Action: s3:PutObject
+#            Resource: !Sub "${S3Bucket.Arn}/*"
+#            Condition:
+#              StringNotEqualsIgnoreCase:
+#                s3:x-amz-grant-full-control: !Sub "id=${SynapseCanonicalId},id=${S3CanonicalUserId}"
+#              StringNotEquals:
+#                s3:x-amz-grant-full-control: !Sub "id=${SynapseCanonicalId},id=${S3CanonicalUserId}"
```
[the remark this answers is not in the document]
zaro0508: i'm not opposed to making that change @clarisse-lau however I'm not sure how your suggestion provides more flexibility?


I'm not sure what your suggestion applies to @xschildw?  is it meant as an alternative to the structure of this `Condition` clause? or is your suggestion meant as a possible work around for the `Deny` policy issue?

templates/htan-synapse-bucket.yaml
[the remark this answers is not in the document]
xdoan: happy to try out `NotPrincipal ` or the multistep upload of owner.txt, though I'd prefer if we could specify this during the ACL just to make it a bit simpler once we get it down.

templates/htan-synapse-bucket.yaml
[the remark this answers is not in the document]
xschildw: My proposal was re: comment to have the Deny policy apply to all principals except one role (the service role that writes owner.txt). To do that we look at 'aws:userid' and exclude everything but roleID:* (so the role would not be denied).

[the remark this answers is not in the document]
xdoan: @zaro0508 this came up because when we were testing (no spaces) one of us had spaces in between the ids and it was not accepted, I think what is in between `" "` needs to be an exact string match.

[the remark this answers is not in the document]
zaro0508: implemented @clarisse-lau and @xschildw suggestions.  Does that look correct @xschildw ?

templates/htan-synapse-bucket.yaml
```diff
@@ -113,14 +118,17 @@ Resources:
             Resource: !Sub "${S3Bucket.Arn}/*"
             Condition:
               StringNotEqualsIgnoreCase:
-                s3:x-amz-grant-full-control: !Sub "id=${SynapseCanonicalId},id=${S3CanonicalUserId}"
+                s3:x-amz-grant-full-control: !Sub "id=${SynapseCanonicalId}, id=${S3CanonicalUserId}"
```
clarisse-lau, line 121: No space between IDs here, so that this string differs from that in the `StringNotEquals` condition and both will be accepted.

zaro0508: done

templates/htan-synapse-bucket.yaml
```diff
@@ -115,12 +120,15 @@ Resources:
               StringNotEqualsIgnoreCase:
                 s3:x-amz-grant-full-control: !Sub "id=${SynapseCanonicalId},id=${S3CanonicalUserId}"
               StringNotEquals:
-                s3:x-amz-grant-full-control: !Sub "id=${SynapseCanonicalId},id=${S3CanonicalUserId}"
+                s3:x-amz-grant-full-control: !Sub "id=${SynapseCanonicalId}, id=${S3CanonicalUserId}"
+              StringNotLike:
+                "aws:userid": !Join ["", [!Ref BucketServiceRoleArn, ":*"]]
```
clarisse-lau, line 125: So the service role ARN can be used in place of the role ID?

templates/htan-synapse-bucket.yaml
```diff
@@ -115,12 +120,15 @@ Resources:
...
 
   # Add owner file to the synapse bucket, requires the cloudformation S3 objects macro
   # https://github.com/Sage-Bionetworks-IT/cfn-s3objects-macro
   SynapseOwnerFile:
     Type: AWS::S3::Object
+    DeoendsOn: "S3BucketPolicy"
```
clarisse-lau, line 131: "DependsOn"

templates/htan-synapse-bucket.yaml
xschildw: @clarisse-lau you're correct, we need the ID.

zaro0508: good catch, fix in here and in depends on PR

templates/htan-synapse-bucket.yaml
zaro0508: fixed

config/prod/htan-bucket-a.yaml
```diff
@@ -17,3 +17,4 @@ parameters:
     - "arn:aws:iam::426577889437:user/clarisse.lau"
   S3CanonicalUserId: "eab4436941f355ce866fcf7944db42020c385ad1f19df8a95704dc4d7552fa06"
   S3SynapseSyncFunctionRoleArn: !stack_output_external "s3-synapse-sync::FunctionRoleArn"
+  BucketServiceRoleArn: !stack_output_external "bootstrap::AWSIAMCfServiceRoleId"
```
clarisse-lau, line 20: Update parameter name

zaro0508: fixed

PR #374, opened 2020-08-26
[IT-937] Setup bucket notification

zaro0508:
Configure bucket notification to trigger lambda function.

### DIFF
--- a/templates/htan-synapse-bucket.yaml
+++ b/templates/htan-synapse-bucket.yaml
@@ -4,6 +4,12 @@ Description: >-
   Synapse S3 Custom Storage
   (https://docs.synapse.org/articles/custom_storage_location.html)
 Parameters:
+  S3SynapseSyncFunctionArn:
+    Type: String
+    Description: The S3 Synapse sync lambda function ARN
+    ConstraintDescription: >-
+      Must be the lambda function ARN
+      (i.e. arn:aws:lambda:us-east-1:787179373106:function:my-lambda-S3SynpaseSyncFunction)
   S3SynapseSyncFunctionRoleArn:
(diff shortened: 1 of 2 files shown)


clarisse-lau:
If the bucket is deployed in a different account from the lambda, permission must be given for it to invoke the lambda (i.e. step 6 in the [setup](https://github.com/Sage-Bionetworks/s3-synapse-sync/blob/master/README.md)) before the notification can be configured. Will that present an issue here?

zaro0508:
> If the bucket is deployed in a different account from the lambda, permission must be given for it to invoke the lambda (i.e. step 6 in the [setup](https://github.com/Sage-Bionetworks/s3-synapse-sync/blob/master/README.md)) before the notification can be configured. Will that present an issue here?

This is the template to create the bucket.  Any bucket that is created from this template will lambda notifications setup (step 6).  When you want to create new buckets (in any account) just use this template to create the bucket and everything is ready to go.

templates/htan-synapse-bucket.yaml
```diff
@@ -64,6 +64,12 @@ Resources:
             AllowedOrigins: ['*']
             AllowedMethods: [GET, POST, PUT, HEAD]
             MaxAge: 3000
+      NotificationConfiguration:
+        LambdaConfigurations:
+          - Event: "s3:ObjectCreated:*"
+            Function: !Ref S3SynapseSyncFunctionRoleArn
```
clarisse-lau, line 70: We need the ARN for the lambda function itself, rather than its role

zaro0508: good catch. fixed

PR #375, opened 2020-08-27
[IT-937] add required AWS lambda permission

zaro0508:
The previous build failed because the NotificationConfiguration requires
a `AWS::Lambda::Permission` which has not been setup yet.  When
attempting to add that permission and NotificationConfiguration at the
same time we run into a circular dependency problem which is documented
in the NotificationConfiguration docs[1].

The doc is basically saying that the only way to avoid the circular
dependency is to deploy in two steps.  The imnplication is that every
time we create a bucket we would need to deploy the AWS::Lambda::Permission
first then run another deploy to setup the NotificationConfiguration.

This step is o setup the permission.

[1] https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-notificationconfig.html

### DIFF
--- a/templates/htan-synapse-bucket.yaml
+++ b/templates/htan-synapse-bucket.yaml
@@ -70,12 +70,22 @@ Resources:
             AllowedOrigins: ['*']
             AllowedMethods: [GET, POST, PUT, HEAD]
             MaxAge: 3000
-      NotificationConfiguration:
-        LambdaConfigurations:
-          - Event: "s3:ObjectCreated:*"
-            Function: !Ref S3SynapseSyncFunctionArn
-          - Event: "s3:ObjectRemoved:*"
-            Function: !Ref S3SynapseSyncFunctionArn
+       # LambdaBucketPermission must be setup before NotificationConfiguration
+       # https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-notificationconfig.html
+#      NotificationConfiguration:
+#        LambdaConfigurations:
+#          - Event: "s3:ObjectCreated:*"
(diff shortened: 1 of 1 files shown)


templates/htan-synapse-bucket.yaml
```diff
@@ -70,12 +70,22 @@ Resources:
...
-            Function: !Ref S3SynapseSyncFunctionArn
+       # LambdaBucketPermission must be setup before NotificationConfiguration
+       # https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-notificationconfig.html
+#      NotificationConfiguration:
+#        LambdaConfigurations:
+#          - Event: "s3:ObjectCreated:*"
+#            Function: !Ref S3SynapseSyncFunctionArn
+#          - Event: "s3:ObjectRemoved:*"
+#            Function: !Ref S3SynapseSyncFunctionArn
+  LambdaBucketPermission:
```
zaro0508, line 81: Adding NotificationConfiguration would fail without LambdaBucketPermission.

I think this can replace the `Sid: "LambdaAccess"` section below.  What do you think @clarisse-lau ?

xschildw: Does not this one give the right to the bucket to invoke the function, and LambdaAccess give the function role the right to get/putObject in the bucket?

clarisse-lau: Yeah, this is the potential issue I was wondering about in my [comment](https://github.com/Sage-Bionetworks/scicomp-provisioner/pull/374#issuecomment-681008691) yesterday. I think we should try this out. But agree with @xschildw , I think we still need the `Sid: "LambdaAccess"` section

PR #378, opened 2020-08-27
[IT-937] redeploy s3-synapse-sync lambda

zaro0508:
Redploy lambda with updates for PR
https://github.com/Sage-Bionetworks/s3-synapse-sync/pull/7


data/s3-synapse-sync-bucket-vars.yaml
```diff
@@ -0,0 +1,5 @@
+s3-synapse-sync-bucket-a:
+  SynapseProjectId: "syn11111"
```
clarisse-lau, line 2: I have this synapse project to test with: syn22335495

zaro0508: updated.

PR #383, opened 2020-08-28
Update bucket name

clarisse-lau:
Update bucket name to match change in https://github.com/Sage-Bionetworks/s3-synapse-sync/pull/13

### DIFF
--- a/config/prod/s3-synapse-sync-bucket-a.yaml
+++ b/config/prod/s3-synapse-sync-bucket-a.yaml
@@ -12,7 +12,7 @@ hooks:
 dependencies:
   - "prod/s3-synapse-sync.yaml"
 parameters:
-  BucketName: "s3-synapse-sync-bucket-a"  # must match bucket name in data/s3-synapse-sync-bucket-vars.yaml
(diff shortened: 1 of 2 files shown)


clarisse-lau:
@zaro0508 @xdoan Please review

config/prod/s3-synapse-sync-bucket-a.yaml
```diff
@@ -12,7 +12,7 @@ hooks:
 dependencies:
   - "prod/s3-synapse-sync.yaml"
 parameters:
-  BucketName: "s3-synapse-sync-bucket-a"  # must match bucket name in data/s3-synapse-sync-bucket-vars.yaml
+  BucketName: "htan-synapse-sync-bucket-a"  # must match bucket name in data/s3-synapse-sync-bucket-vars.yaml
```
zaro0508, line 15: as you probably know.  AWS does not allow changing a bucket name once it's been provisioned.  You probably want to create a new bucket and then delete  `htan-synapse-sync-bucket-a` bucket if you do not need it anymore.

clarisse-lau: Right, we'll create a new bucket

PR #390, opened 2020-09-02
create new htan test bucket

clarisse-lau:
Create new s3-synapse-sync bucket, adding service account `SynapseID`

@zaro0508 @xdoan

### DIFF
--- a/config/prod/htan-dcc-centerA.yaml
+++ b/config/prod/htan-dcc-centerA.yaml
@@ -1,0 +1,30 @@
+# Provision a Synapse External Bucket (http://docs.synapse.org/articles/custom_storage_location.html) that will
+# work with the s3-synapse-sync lambda: https://github.com/Sage-Bionetworks/s3-synapse-sync
+template_path: "remote/s3-synapse-sync-bucket.j2"
+stack_name: "htan-dcc-centerA"
+stack_tags:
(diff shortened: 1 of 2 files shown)


config/prod/htan-dcc-centerA.yaml
```diff
@@ -0,0 +1,30 @@
...
+dependencies:
+  - "prod/s3-synapse-sync.yaml"
+parameters:
+  BucketName: "htan-dcc-centerA"  # must match bucket name in data/s3-synapse-sync-bucket-vars.yaml
+  SynapseIDs:
+    - "3382314"
+    - "3408101"
+    - "3413795"
+  S3UserARNs:
+    - "arn:aws:sts::563295687221:assumed-role/sandbox-developer/xengie.doan@sagebase.org"
```
zaro0508, line 21: 3 synapse IDs but only 2 S3UserARNs?  should there be another S3UserArn?

clarisse-lau: yes, just 2 S3UserARNs. We got this error when testing an upload `403 Client Error: 
Only the creator of ExternalS3StorageLocationSetting.id=41090 can create an external S3FileHandle with storageLocationId = 41090` and think it may be because the Synapse service account profile is not included in owner.txt

PR #401, opened 2020-09-11
update bu folders

clarisse-lau:
- update BU folder list
- remove `htan-dcc-center-b` bucket


xschildw (approved):
Note: there are more folders in the list than there are on Synapse in syn22124336, not sure if that matters.

clarisse-lau:
Thanks @xschildw, this is correct. The lambda will create these folders on Synapse when filehandles are added

PR #425, opened 2021-01-15
Remove offboarded user's instance

ahayden:
This instance was left running and needs to be terminated and stack deleted.


zaro0508 (approved):

ahayden:
Want to make sure you see that I added some clean up of old stacks in a follow up commit.

xschildw (approved):

PR #435, opened 2021-03-05
Remove Kara's instances and permissions

karawoo:
Removes two EC2 instances I created (these were for interactive work, not running anything ongoing) and removes me from the data usages statistics S3 bucket


zaro0508 (approved):

config/prod/usage-statistics-S3.yaml
```diff
@@ -10,7 +10,6 @@ parameters:
   OwnerEmail: 'larsson.omberg@sagebase.org'
```
zaro0508, line 10: should this change to @tthyer ?

zaro0508 (changes requested):

config/prod/usage-statistics-S3.yaml
tthyer: No, let's leave him as owner of that (it ain't broke)
I'll make a second pass on this and fix some things up in a separate PR

zaro0508 (approved):

PR #440, opened 2021-03-24
pin s3-synapse-sync v1.3.0

clarisse-lau:


config/prod/s3-synapse-sync.yaml
```diff
@@ -8,7 +8,7 @@ dependencies:
   - "prod/htan-synapse-sync-kms-key.yaml"
 hooks:
   before_launch:
-    - !cmd "curl https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/s3-synapse-sync/master/s3-synapse-sync.yaml --create-dirs -o templates/remote/s3-synapse-sync.yaml"
+    - !cmd "curl https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/s3-synapse-sync/v1.1/s3-synapse-sync.yaml --create-dirs -o templates/remote/s3-synapse-sync.yaml"
```
zaro0508, line 11: Your tag doesn't seem to match the version that's in the cfn template, https://github.com/Sage-Bionetworks/s3-synapse-sync/blob/master/template.yaml#L19

It would be better if it matches.  You tag seems to be ok though, so either change the version references in the template or redo your tags. 

here's the deployed file from your tag, https://bootstrap-awss3cloudformationbucket-19qromfd235z9.s3.amazonaws.com/index.html#s3-synapse-sync/v1.1/

zaro0508 (changes requested):
the PR is ok, but the tags are wacked.

zaro0508:
The PR looks fine however your tag didn't trigger a build, https://travis-ci.com/github/Sage-Bionetworks/s3-synapse-sync/builds which means no deploy to s3 bucket, https://bootstrap-awss3cloudformationbucket-19qromfd235z9.s3.amazonaws.com/index.html#s3-synapse-sync/

here's the matcher that will trigger a build, https://github.com/Sage-Bionetworks/s3-synapse-sync/blob/master/.travis.yml#L19

i think the tag needs to specifically be `v1.3.0`

zaro0508 (approved):

zaro0508 (changes requested):
No this is wrong.  you want to pin to a version that works with existing HTAN deployments so we don't break anything.  Was that v1.0 or v1.1?

clarisse-lau:
v1.1, though the only thing that has changed between v1.1 to v1.3.0 is the version referenced in template.yaml?

zaro0508 (approved):
ahh you are correct. i forgot those changes have not merged yet.  ok, looks good.

PR #454, opened 2021-04-12
Remove Megha's EC2

zaro0508:
He is switching to use the service catalog.


itismeghasyam:
Khai, I went in and collected files that I would need.

You can go ahead and remove this EC2 instance.

itismeghasyam (approved):

ahayden (approved):
Thanks @itismeghasyam !

PR #455, opened 2021-04-16
Remove Will's instance

zaro0508:
We are deprecating scicomp as the self-service provisioning system.
Everyone should transition to using the Sage service catalog to
get cloud resources going foward.


zaro0508:
@wpoehlm please approve when ready for us to remove your resource

wpoehlm (approved):

PR #458, opened 2021-04-26
Add buckets for DIAN data migration sub-project

ahayden:
This change adds two buckets using the standard template. One (`dian-hm-uat-migration-s3`) is for validating the data transfer process. The other (`dian-hm-prod-migration-s3`) is for production data migration. These data will be ingested and pushed into Synapse projects. Each bucket will need write access by an ARN provided by Happy Medium from their uat and prod tenants.
- [x] Clearly explain your change with a descriptive commit message
- [x] Passes pre-commit

PR #461, opened 2021-05-07
Peg deployed Synapse-attached bucket to v0.2.5 aws-infra tag

ahayden:
- [x] This change pegs the only deployed Synapse-attached S3 bucket to v0.2.5 of aws-infra so that https://github.com/Sage-Bionetworks/aws-infra/pull/284/ can be merged without risk of breaking the deployed bucket connection
- [x] Passes pre-commit

https://github.com/Sage-Bionetworks/aws-infra/pull/284 depends on this if we want to merge it to unblock https://github.com/Sage-Bionetworks/scicomp-provisioner/pull/458

### DIFF
--- a/config/prod/TESLA-download-bucket.yaml
+++ b/config/prod/TESLA-download-bucket.yaml
@@ -44,6 +44,6 @@ parameters:
 # For CI system (do not change)
 hooks:
   before_launch:
-    - !cmd "curl https://{{stack_group_config.admincentral_cf_bucket}}.s3.us-east-1.amazonaws.com/aws-infra/v0.2.5/s3-bucket-v2.j2 --create-dirs -o templates/remote/s3-bucket-v2.j2"
+    - !cmd "curl https://{{stack_group_config.admincentral_cf_bucket}}.s3.us-east-1.amazonaws.com/aws-infra/v0.2.8/s3-bucket-v2.j2 --create-dirs -o templates/remote/s3-bucket-v2.j2"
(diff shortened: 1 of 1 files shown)


config/prod/TESLA-download-bucket.yaml
```diff
@@ -44,6 +44,6 @@ parameters:
 # For CI system (do not change)
 hooks:
   before_launch:
-    - !cmd "curl https://{{stack_group_config.admincentral_cf_bucket}}.s3.us-east-1.amazonaws.com/aws-infra/master/s3-bucket-v2.j2 --create-dirs -o templates/remote/s3-bucket-v2.j2"
+    - !cmd "curl https://{{stack_group_config.admincentral_cf_bucket}}.s3.us-east-1.amazonaws.com/aws-infra/v0.2.5/s3-bucket-v2.j2 --create-dirs -o templates/remote/s3-bucket-v2.j2"
```
zaro0508, line 47: there's already a `v0.2.7` so let's set this to `v0.2.8` after we pin to the current version by merging https://github.com/Sage-Bionetworks/sandbox-provisioner/pull/385 and #462

zaro0508 (changes requested):

zaro0508:
actually PR #462 pins all references to templates in aws-infra therefore this one is not needed?

ahayden:
True- we don't actually need to pin all buckets. But, let's use #462 instead

PR #472, opened 2021-06-02
update reesouce access fpr SSO

zaro0508:
We switched from SAML integration to SSO which uses a different role
for federated user access.  We update with the new SSO role.


brucehoff (approved):
We also need to add Nicole Kauer to AMP-Emory-Bucket, but we can do that in a subsequent PR.

PR #475, opened 2021-06-04
Replace SAML reference with SSO

zaro0508:
We've switched from using JC AWS SAML integration to AWS SSO so
we need to update the reference to the role.


zaro0508:
merging to fix the build.

PR #478, opened 2021-06-07
Provision miniDREAM 2021 infrastructure

BrunoGrandePhD:
I'm deploying infrastructure for miniDREAM 2021. This stack is composed of a high-memory EC2 instance in a private subnet behind by a public load balancer (both using existing aws-infra templates). 

We will be hosting RStudio on port 8787 of this instance, and we will setup an nginx reverse proxy to redirect incoming HTTPS (port 443) traffic from the ALB to port 8787. The ALB will take care of upgrading HTTP (port 80) traffic to HTTPS (port 443). 

We have a custom setup to integrate with Synapse evaluation queues, so we can't use the RStudio packer AMI.

### DIFF
--- a/config/prod/minidream-rstudio-2021-ec2-alb.yaml
+++ b/config/prod/minidream-rstudio-2021-ec2-alb.yaml
@@ -1,0 +1,22 @@
+# Provision a public load balancer in front of private miniDREAM EC2
+# Based on prod/minidream-R-env-pub-prod-instance-ec2-alb-v2.yaml
+template_path: "remote/managed-alb-https-v2.yaml"
+stack_name: "minidream-rstudio-2021-ec2-alb"
+
+parameters:
+  Department: "CompOnc"
+  Project: "CSBC"
+  OwnerEmail: "bruno.grande@sagebase.org"
+  Ec2InstanceId: !stack_output "prod/minidream-rstudio-2021-ec2.yaml::Ec2InstanceId"
+  SSLCertificateIdArn: "arn:aws:acm:us-east-1:055273631518:certificate/992c2d49-579a-465e-beff-1a08c284f7db"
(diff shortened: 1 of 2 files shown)


thomasyu888 (approved):

config/prod/minidream-rstudio-2021-ec2-alb.yaml
```diff
@@ -0,0 +1,22 @@
+# Provision a public load balancer in front of private miniDREAM EC2
+# Based on prod/minidream-R-env-pub-prod-instance-ec2-alb-v2.yaml
+template_path: "remote/managed-alb-https-v2.yaml"
+stack_name: "minidream-rstudio-2021-ec2-alb"
+
+parameters:
+  Department: "CompOnc"
+  Project: "CSBC"
+  OwnerEmail: "bruno.grande@sagebase.org"
+  Ec2InstanceId: !stack_output "prod/minidream-rstudio-2021-ec2.yaml::Ec2InstanceId"
```
zaro0508, line 10: this depends on the deployment of `config/prod/minidream-rstudio-2021-ec2.yaml`.  i think you need to add a `dependencies` to this file so that this template will get deployed after the EC2 template.  Something like this.. https://github.com/Sage-Bionetworks/scicomp-provisioner/blob/a7a43e4b4221bba40a0e6dd4dacfad4464e140f6/config/prod/htan-dcc-dfci.yaml#L12

zaro0508 (changes requested):

zaro0508 (approved):

config/prod/minidream-rstudio-2021-ec2-alb.yaml
BrunoGrandePhD: image

PR #479, opened 2021-06-07
fix sceptre dependency

zaro0508:


BrunoGrandePhD (approved):
Let's see if this is the issue or if I misinterpreted the syntax from the [docs](https://sceptre.cloudreach.com/2.3.0/docs/resolvers.html#stack-output). Namely, I don't know if the `shared/` part in their example value corresponds to our `prod/` folder. 

Alternatively, we can just populate the instance ID (`i-061c2c5354f379648`) or try `stack_output_external` instead.

PR #482, opened 2021-06-22
[IT-1295] fix travis deployment

zaro0508:
Travis is cacheing templates which can sometimes cause resources
to provision with old templates.  We delete the foler containing
downloaded templates to force travis to downlod them on every
deployment.


xschildw (approved):

brucehoff (approved):
Good sleuthing, @zaro0508 !

PR #487, opened 2021-06-25
[IT-1295] Fix sceptre template refernces

zaro0508:
wget when used without the `-O` option will not download
a file if it already exists.  Since we have multiple
sceptre templates referencing different versions of a CFN
template it can download the 1st version then not download the
2nd version.

For example:

assume the following:
* config/prod/a.yaml references v1.0 of x.yaml CFN template
* config/prod/b.yaml references v2.0 of x.yaml CFN template

workflow:
* sceptre runs config/prod/a.yaml and downloads v1.0 of x.yaml
* sceptre runs config/prod/b.yaml but does not download v2.0 of x.yaml
becuase v1.0 of x.yaml already exists.  sceptre will deploy with the old v1.0 of x.yaml

To force wget to download and overwrite the file we need to use
the `-O` option.

### DIFF
--- a/config/prod/dian-prod-hm-migration-s3.yaml
+++ b/config/prod/dian-prod-hm-migration-s3.yaml
@@ -45,4 +45,4 @@ parameters:
 # For CI system (do not change)
 hooks:
   before_launch:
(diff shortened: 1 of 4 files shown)


zaro0508:
@tthyer @BrunoGrandePhD  and @thomasyu888 if you are using wget consider updating your sceptre repos with this fix as well.

brucehoff (approved):
Good sleuthing!

PR #504, opened 2021-12-07
Terminate miniDREAM server and load balancer

BrunoGrandePhD:
We no longer need this miniDREAM EC2 instance and associated load balancer. They can be terminated now. I've backed up everything on the server to GitHub, so these stacks can be deleted whenever is convenient for IT.


zaro0508:
thanks @BrunoGrandePhD

zaro0508 (approved):

PR #508, opened 2022-01-20
[IT-1551] Apply CostCenter tags to resources

zaro0508:
Add a [Sage CostCenter tag](https://github.com/Sage-Bionetworks/aws-infra/tree/master/templates/tags)
to resources deployed by the scicomp provisioner using
[sceptre stack_tags](https://docs.sceptre-project.org/dev/docs/stack_config.html#stack-tags)

Also change sceptre configs to use sceptre
[template handlers](https://docs.sceptre-project.org/dev/docs/template_handlers.html)
instead of hooks.

depends on https://github.com/Sage-Bionetworks/aws-infra/pull/332


brucehoff:
This is actually related to IT-1551 (SciComp) not IT-1552 (Sandbox).

config/prod/JKG-SciComp-S3.yaml
```diff
@@ -1,6 +1,13 @@
...
-template_path: "remote/s3-bucket.yaml"
+template:
+  type: "http"
+  url: "https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/v0.2.7/s3-bucket.yaml"
 stack_name: "JKG-SciComp-S3"
+stack_tags:
+  Department: "SysBio"
+  Project: "amp-ad"
+  OwnerEmail: 'jake.gockley@sagebase.org'
+  CostCenter: "NIA AMP-AD CC / 101500"
```
brucehoff, line 10: Added @jgockley62  as a reviewer, to confirm this CostCenter.

config/prod/TESLA-download-bucket.yaml
```diff
@@ -1,7 +1,14 @@
...
-template_path: "remote/s3-bucket-v2.j2"
+template:
+  type: "http"
+  url: "https://{{stack_group_config.admincentral_cf_bucket}}.s3.us-east-1.amazonaws.com/aws-infra/v0.2.7/s3-bucket-v2.j2"
 stack_name: "TESLA-download-bucket"
+stack_tags:
+  Department: "CompOnc"
+  Project: "TESLA"
+  OwnerEmail: "james.eddy@sagebase.org"
+  CostCenter: "Other / 000001"
```
brucehoff, line 11: Since `CostCenter` is "Other", there should be an OtherCostCenter (or similar, I forget the exact tag name) with the tag for TESLA.  Remember that if there *is* no cost center (e.g. no specific project), then it's OK to have "No Program / 000000".

config/prod/mep-lincs-s3.yaml
```diff
@@ -3,8 +3,15 @@
...
-template_path: "remote/s3-bucket.yaml"
+template:
+  type: "http"
+  url: "https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/v0.2.7/s3-bucket.yaml"
 stack_name: "mep-lincs-s3"
+stack_tags:
+  Department: "SysBio"
+  Project: "MEP-LINCS"
+  OwnerEmail: "larsson.omberg@sagebase.org"
+  CostCenter: "Other / 000001"
```
brucehoff, line 14: Where's the "Other" cost center?

config/prod/usage-statistics-S3.yaml
```diff
@@ -1,6 +1,14 @@
...
-template_path: "remote/s3-bucket.yaml"
+template:
+  type: "http"
+  url: "https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/v0.2.7/s3-bucket.yaml"
 stack_name: "usage-statistics-S3"
+stack_tags:
+  Department: "SysBio"
+  Project: "DataEngineering"
+  OwnerEmail: 'larsson.omberg@sagebase.org'
+  CostCenter: "Other / 000001"
```
brucehoff, line 10: Where's the "Other" cost center?

config/prod/veoibd-s3.yaml
```diff
@@ -1,7 +1,13 @@
...
 # http://docs.synapse.org/articles/custom_storage_location.html
-template_path: "remote/s3-bucket.yaml"
+template:
+  type: "http"
+  url: "https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/v0.2.7/s3-bucket.yaml"
 stack_name: "veoibd-s3"
+stack_tags:
+  Department: "SysBio"
+  Project: "VEO-IBD"
+  OwnerEmail: "nicole.kauer@sagebase.org"
```
brucehoff, line 10: No `CostCenter`?  Remember we have a "No Program" value for resources that are not directly funded.

brucehoff (changes requested):

config/prod/TESLA-download-bucket.yaml
zaro0508: it was just a guess, do you have a better suggestion @brucehoff ?  otherwise maybe @jaeddy can suggest a more appropriate value?

config/prod/veoibd-s3.yaml
zaro0508: oops.  i'm assuming the OwnerEmail should change for this as well since nicole is no longer here.  Who would know the most appropriate person to own this now?  maybe they can also suggest a costcenter tag value.

config/prod/mep-lincs-s3.yaml
zaro0508: it's here https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/tags/SageFinancialProgramCodes.json#L19

but this is another guess,  @larssono would you be able to suggest a cost center tag or this? options are here: https://github.com/Sage-Bionetworks/aws-infra/tree/master/templates/tags

config/prod/usage-statistics-S3.yaml
zaro0508: it's here https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/tags/SageFinancialProgramCodes.json#L19

but this is another guess,  @larssono would you be able to suggest a cost center tag or this? options are here: https://github.com/Sage-Bionetworks/aws-infra/tree/master/templates/tags

config/prod/mep-lincs-s3.yaml
brucehoff: @zaro0508 I'm not asking "where is the list of program codes."  I'm asking, "I thought we had the convention that if the the tag 'CostCenter' == 'Other' then there must be a second tag 'CostCenterOther' with the actual cost center.  Why doesn't this PR follow the convention?"

zaro0508: @brucehoff We only used that convention when provisioning with the service catalog due to the 25 tag option limitation.  This is the PR provisioner system which does not have that limitation so we don't need to use that convention here.

brucehoff: @zaro0508 I thought the only reason we use "Other / 000001" is to indicate that the actual cost center is in the "CostCenterOther" tag.  If so, we shouldn't use "Other / 000001" here, we should use the actual program code or "NO PROGRAM / 00000".  If I am misunderstanding, let me know.

config/prod/veoibd-s3.yaml
thomasyu888: @zaro0508 put me first as maintainer then.  CostCenter can probably be determined by Larsson.

config/prod/usage-statistics-S3.yaml
brucehoff: If @larssono is not responsive, maybe just use No Program / 00000.

config/prod/JKG-SciComp-S3.yaml
jgockley62: @brucehoff not familiar with the new cost center tags, but this is for AMP-AD

jgockley62 (approved):

config/prod/JKG-SciComp-S3.yaml
brucehoff: @jgockley62 If you haven't been provided a specific cost center than the next step would be for you to ask the project lead/PI what program funds the work. Finance maintains a list of six digit program codes, and the project leads/PIs should know where their funding is coming from.

brucehoff: @zaro0508 Since Jake has not suggested a different code, I reviewed the available codes and I think this one ("NIA AMP-AD CC / 101500") is the best we can guess.

config/prod/TESLA-download-bucket.yaml
brucehoff: @zaro0508 I know of no program code for Tesla so please use "NO PROGRAM / 000000"

config/prod/veoibd-s3.yaml
zaro0508: changed OwnerEmail to @thomasyu888 and set CostCenter to "NO PROGRAM / 000000"

PR #513, opened 2022-02-24
Enable Stanford S3 bucket notifs

clarisse-lau:
Configuring the HTAN Stanford bucket with synapse sync lambda, and granting access to collaborator Daniel Cotter

### DIFF
--- a/config/prod/htan-dcc-stanford.yaml
+++ b/config/prod/htan-dcc-stanford.yaml
@@ -20,11 +20,13 @@ parameters:
     - "arn:aws:sts::563295687221:assumed-role/AWSReservedSSO_Developer_baa6fed639faf5e7/adam.taylor@sagebase.org"
     - "arn:aws:iam::728882028485:role/htan-project-TowerForgeBatchHeadJobRole-17MIZHWBA1TC5"
     - "arn:aws:iam::728882028485:role/htan-project-TowerForgeBatchWorkJobRole-17360PQARGWYU"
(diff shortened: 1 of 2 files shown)


BrunoGrandePhD:
@clarisse-lau: Out of curiosity, why was this PR closed? I wasn't yet involved in this discussion until afterwards, so I want to make sure I understand the context behind the decision in case I don't have the full picture.

clarisse-lau:
@BrunoGrandePhD No, thanks for reopening. 
I may be the one without the full picture, as we haven’t yet had a discussion re: Globus at the DCC calls. I had closed it because with the resolution of the Jira ticket, it seemed that this required a larger DCC/SC discussion and we'd be holding off until that occurred; I wasn’t aware you were already spec-ing out requirements for use with the HTAN buckets

BrunoGrandePhD:
I think your instinct with this PR was correct. I don't think a larger DCC/SC discussion needs to happen in this case because the Globus S3 connector appears to act like any other S3 client (_e.g._ the AWS CLI). It remains to be confirmed whether it will work as such, but if that's indeed the case, I think this PR might be enough to grant them access to the bucket. 

Let's wait for Karl to confirm that they will follow the instructions that I laid out in my email before merging this though.

BrunoGrandePhD (approved):
With Karl's email, I think we can proceed with merging this PR. 

@zaro0508: Review, please? Thanks!

zaro0508 (approved):

PR #518, opened 2022-04-11
[IT-1748] fix EC2

zaro0508:
This fixes an EC2 that was provisioned with a bad CFN template
(`EC2-gene-co-expression.yaml`).  Once verified we will remove
EC2 from `EC2-gene-co-expression.yaml` template.


jgockley62:
Thanks!

jgockley62 (commented):
Thanks!

PR #530, opened 2022-07-14
Switch to scicomp role for administering HTAN buckets

BrunoGrandePhD:
This PR is partially motivated by IT-1967 and partially motivated by the need to inspect the bucket policy, which can only be done by IAM identities within the same account as the S3 bucket. As a result, my current sandbox role doesn't have access to the bucket policy. Since I have access to the viewer role in scicomp, I was thinking of granting access to my specific assumed-role at the resource/bucket level. If this works, then we can migrate others to using the same viewer role to administer/access the HTAN buckets.

### DIFF
--- a/config/prod/htan-dcc-center-a.yaml
+++ b/config/prod/htan-dcc-center-a.yaml
@@ -31,7 +31,7 @@ parameters:
     - "arn:aws:iam::728882028485:role/htan-project-TowerForgeBatchWorkJobRole-17360PQARGWYU"
   S3AdminARNs:
     - "arn:aws:iam::426577889437:user/clarisse.lau"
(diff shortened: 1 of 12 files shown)


jesusaurus (approved):
🤞

zaro0508 (approved):
just to be clear, you are actually giving access to your [viewer-plus role](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/700-aws-sso/_tasks.yaml#L334-L358) @BrunoGrandePhD.  In any case this looks fine and we can probably give other sage people viewer plus access to scicomp.

BrunoGrandePhD:
Feel free to merge this PR. I'll test it to make sure it works. If it does, we can open a separate PR to propagate the switch to all other sandbox roles.

PR #539, opened 2022-10-06
Create recover-s3-bucket.yaml

itismeghasyam:
- Creating an S3 bucket for the RECOVER project for Care Evolution to export data into
- To validate the files, I ran `pre-commit run --all-files` successfully


brucehoff:
I don't think we want to provision this bucket via SciComp because then SciComp would be in our FISMA boundary.  It may be better to provision a separate AWS account for Recover.

zaro0508:
@itismeghasyam @scarlson-sage RECOVER cost center will be add in PR https://github.com/Sage-Bionetworks/aws-infra/pull/358.  Please review and if it looks good to you, approve it.

config/prod/recover-s3-bucket.yaml
```diff
@@ -0,0 +1,24 @@
+# Provision a Synapse External Bucket (http://docs.synapse.org/articles/custom_storage_location.html)
+template:
+  type: "http"
+  url: "https://raw.githubusercontent.com/Sage-Bionetworks/aws-infra/v0.4.8/templates/S3/synapse-external-bucket.j2"
+stack_name: "recover-s3-bucket"
+stack_tags:
+  Department: "SysBio"
+  Project: "recover"
```
zaro0508, line 8: should this be `RECOVER`?

zaro0508 (changes requested):

config/prod/recover-s3-bucket.yaml
```diff
@@ -0,0 +1,24 @@
...
+  # The following parameters are only examples they are not required.
+  # You may omit them if you do not need to override the defaults.
+  # (Optional) true (default) to encrypt bucket, false for no encryption
+  # (Optional) true for read-write bucket, false (default) for read-only bucket
+  AllowWriteBucket: 'true'
+  # (Optional) Synapse username (default: ""), required if AllowWriteBucket=true
+  # (Optional) Allow accounts, groups, and users to access bucket (default is no access).
+  GrantAccess:
+    - 'arn:aws:iam::325565585839:root'   # Required ARN for a synapse bucket
+    - 'arn:aws:iam::111111111111:role/EES3-RK-AAAAAA-Recover' # Care Evolution access
```
zaro0508, line 21: this does not look like a valid ARN.  you need a valid ARN otherwise AWS will fail to provision this bucket.  If you don't have a valid ARN yet you should remove this for now and then update this PR with a valid ARN at a later time.

config/prod/recover-s3-bucket.yaml
zaro0508: AWS uses that tag to query and run reports so it might be good for consistency purposes.

config/prod/recover-s3-bucket.yaml
```diff
@@ -0,0 +1,24 @@
...
+  CostCenter: "MGH RECOVER / 122500"
+parameters:
+  # The following parameters are only examples they are not required.
+  # You may omit them if you do not need to override the defaults.
+  # (Optional) true (default) to encrypt bucket, false for no encryption
+  # (Optional) true for read-write bucket, false (default) for read-only bucket
+  AllowWriteBucket: 'true'
+  # (Optional) Synapse username (default: ""), required if AllowWriteBucket=true
+  # (Optional) Allow accounts, groups, and users to access bucket (default is no access).
+  GrantAccess:
```
zaro0508, line 19: forgot one more thing here.  if you need access to this bucket you'll want to add an ARN for yourself.  so adding this ARN will do it.. `arn:aws:sts::526515999252:assumed-role/AWSReservedSSO_S3ExternalCollab_40c062f682e7f3f5/meghasyam@sagebase.org`

zaro0508 (changes requested):

config/prod/recover-s3-bucket.yaml
itismeghasyam: 👍Will change to 'RECOVER'

config/prod/recover-s3-bucket.yaml
itismeghasyam: Thanks Khai! Will add this too

zaro0508 (approved):

PR #555, opened 2022-11-09
Update nf-syn23664726-s3.yaml

allaway:
Adding a Tower user to be able to run [nf-synindex](https://github.com/Sage-Bionetworks-Workflows/nf-synindex) on the bucket.

### DIFF
--- a/config/prod/nf-syn23664726-s3.yaml
+++ b/config/prod/nf-syn23664726-s3.yaml
@@ -20,7 +20,6 @@ parameters:
     - 'arn:aws:iam::325565585839:root'   # Required ARN for a synapse bucket
     - 'arn:aws:sts::526515999252:assumed-role/AWSReservedSSO_S3ExternalCollab_40c062f682e7f3f5/robert.allaway@sagebase.org'
     - 'arn:aws:iam::694704239947:root' ## Nebula Genomics access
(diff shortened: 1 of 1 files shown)


config/prod/nf-syn23664726-s3.yaml
```diff
@@ -21,6 +21,8 @@ parameters:
     - 'arn:aws:sts::526515999252:assumed-role/AWSReservedSSO_S3ExternalCollab_40c062f682e7f3f5/robert.allaway@sagebase.org'
     - 'arn:aws:iam::694704239947:root' ## Nebula Genomics access
     - 'arn:aws:iam::694704239947:user/giri' ## Nebula Genomics access
```
zaro0508, line 23: should we remove one of these from the list?

zaro0508 (approved):

config/prod/nf-syn23664726-s3.yaml
allaway:
```suggestion
    - 'arn:aws:iam::694704239947:root' ## Nebula Genomics access
```

allaway: Yes! Done.

allaway:
I found that one of 3k files was missing, so I would prefer not to mess with permissions until they xfer the final file. :) Will reopen when ready.

PR #557, opened 2022-11-17
Refactor CI deployment for multiple regions

zaro0508:
We now have a need to support deployments to multiple AWS regions. Change to the CI system to accomodate multi-region deployments.

### DIFF
--- a/config/prod/usage-statistics-S3.yaml
+++ b/config/prod/usage-statistics-S3.yaml
@@ -1,7 +1,7 @@
 # For storage of files related to usage statistics reporting
 template:
   type: "http"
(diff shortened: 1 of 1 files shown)


zaro0508:
some of the templates have dependencies on lambdas which are region specific.  This means that we would need to deploy the lambda into every supported region as well.

config/ue-central-1/config.yaml
```diff
@@ -0,0 +1 @@
+region: {{ var.region | default("eu-central-1") }}
```
jesusaurus, line 1: typo in the directory name here, otherwise this PR looks good

PR #561, opened 2023-02-22
Disable ACLs on usage-statistics bucket

zaro0508:
There were issues with access to the usage-statistics bucket due to ACLs being enabled on that bucket.  AWS now advises using bucket policies for access instead of ACLs[1].  Our newer cloudformation template deploys S3 with ACLs turned off. This configures the bucket with settings from our updated cloudformation template.

[1] https://docs.aws.amazon.com/AmazonS3/latest/userguide/security-best-practices.html


avanlinden (approved):
Looks good to me

PR #564, opened 2023-03-17
Recover

itismeghasyam:
**PR Checklist:**

_S3 Buckets for RECOVER Project:_

1. (First Bucket) _recover_input_shared_bucket_: We receive data from Care Evolution in this bucket.
2. (Second Bucket) _recover_synapse_attached_bucket_: This would be a copy of the first bucket (i.e data received in the first bucket would regularly be copied into this) and only Sage would have access to this bucket. This would also be linked to a Synapse project (RECOVER -  syn51105296) using an STS token.


thomasyu888:
@itismeghasyam Thanks for doing this.  First thing, can you separate out each of the files into their own PRs so we can comment and work through each one?   This will allow us to separate out the concerns - sorry for the inconvenience!

PR #565, opened 2023-03-17
S3 Buckets for RECOVER Project, Bucket 2: recover_synapse_attached_bucket

itismeghasyam:
2. (Second Bucket) recover_synapse_attached_bucket: This would be a copy of the first bucket (i.e data received in the first bucket would regularly be copied into this) and only Sage would have access to this bucket. This would also be linked to a Synapse project using an STS token.


itismeghasyam:
PR #564 was split into two, this is one and the other is #566 

Changes from #564 -> Owner's email updated to the shared RECOVER account - 'synapse-service-sysbio-recover-data-storage-01@sagebase.org' from 'meghasyam@sageabase.org'

config/prod/recover-synapse-attached-input-bucket.yaml
```diff
@@ -0,0 +1,42 @@
...
+  Department: "SysBio"
+  # The Sage project (Infrastructure, amp-ad, ntap, dream, etc.,)
+  Project: "RECOVER"
+  # The resouce owner
+  OwnerEmail: 'synapse-service-sysbio-recover-data-storage-01@sagebase.org'
+  # This is the shared account (credentials shared on Last Pass) for RECOVER
+  # (Optional) true for read-write bucket, false (default) for read-only bucket
+  AllowWriteBucket: 'true'
+  # (Optional) Synapse username (default: ""), required if AllowWriteBucket=true
+  SynapseUserName: 'meghasyam'
```
thomasyu888, line 35: After thinking a little about this.  Were you thinking of actually creating entities in Synapse or doing something similar to what we do for the parquet datasets in BridgeDowstream (access via STS)?

The reason I ask is because `SynapseUserName` is specified for if you want to create file entities in Synapse that link to S3 objects in this bucket. (I have the question for the other bucket as well). Happy to hop on a quick slack huddle Monday.

PR #570, opened 2023-03-27
Update pre-commit hook to new endpoint

jesusaurus:
The mips-api lambda is migrating to a new host name.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
     -    id: check-stack-names
     -    id: check-stack-tag-values
          args: [--tag=CostCenter,
-                --file=https://finops-api.sageit.org/tags,
+                --file=https://mips-api.finops.sageit.org/tags,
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -24,7 +24,7 @@ repos:
     -    id: check-stack-names
     -    id: check-stack-tag-values
          args: [--tag=CostCenter,
-                --file=https://finops-api.sageit.org/tags,
+                --file=https://mips-api.finops.sageit.org/tags,
```
zaro0508, line 27: the new endpoint doesn't work.  is this dependent on another PR somewhere?  if so can you please reference it?

zaro0508 (changes requested):

.pre-commit-config.yaml
jesusaurus: I was expecting this to work after a successful run of aws-deploy, but it looks like the first run of the lambda hit a 500 during the log in and got locked out for an hour. It's working now.

PR #578, opened 2023-04-14
Remove region restriction for TESLA bucket

zaro0508:
The template used to provision the TESLA bucket is old and is setup to deploy a lambda with python 3.6. AWS no longer supports that old python version so it fails when we attempt to deploy the bucket with the following message..

```
Resource handler returned message: "The runtime parameter of python3.6 is no longer supported
for creating or updating AWS Lambda functions. We recommend you use the new runtime (python3.9)
while creating or updating functions.
```

We need to temporarily remove region restriction, update the templates then hopefully restore region restriction.


xschildw (approved):
Is there an issue to upgrade the lambda / re-instate the restriction?

zaro0508:
@xschildw yes, that's what we need to do, check issue IT-2794

PR #582, opened 2023-04-21
Remove unneeded parameter

jesusaurus:
The mips-api lambda no longer includes an Other tag by default, so we don't need to ignore it.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,7 +26,6 @@ repos:
          args: [--tag=CostCenter,
                 --file=https://mips-api.finops.sageit.org/tags,
                 --file=DeprecatedProgramCodes.json,
-                --exclude=Other / 000001
          ]
 -   repo: https://github.com/sirosen/check-jsonschema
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -26,7 +26,6 @@ repos:
          args: [--tag=CostCenter,
                 --file=https://mips-api.finops.sageit.org/tags,
```
zaro0508, line 27: i think you'll need to update this url to include the `Other` tag

zaro0508 (changes requested):

.pre-commit-config.yaml
jesusaurus: are you sure? we have been excluding the value as a valid tag value, now that it's not even listed in the default output why not just remove the exclude?

zaro0508: ahh yes, i think you are correct.

zaro0508 (approved):

PR #590, opened 2023-06-02
Update recover-s3-bucket.yaml

itismeghasyam:
Remove Care Evolution access from the Pilot bucket to ensure that production data does not get uploaded here by mistake.


zaro0508 (approved):

itismeghasyam:
@zaro0508 Can you revert these changes and reinstate Care Evolution access to the pilot testing data bucket.

We have new data format from a different wearable (Garmin) in the RECOVER study, and want to take a look at the data before it is pushed into prod bucket.

PR #591, opened 2023-06-06
[IT-2859 & IT-1505] Enable bucket life cycle

zaro0508:
Enable bucket life cycle for buckets:
* amp-mayo-sinai-synapseencryptedexternalbucket-1bmvn8rlwixv2
* nf-syn23664726-s3-bucket-n9uakf7bowwd
* gates-ki-001-synapseencryptedexternalbucket-kl8vu7tqto1b

### DIFF
--- a/config/prod/amp-mayo-sinai.yaml
+++ b/config/prod/amp-mayo-sinai.yaml
@@ -31,6 +31,6 @@ parameters:
   # to download objects in this bucket when not using an AWS resource in the same region!
   SameRegionResourceAccessToBucket: "true"
 
-  # Enable Intelligent-tiering storage
+  # Enable transition to deep archive storage
   EnableDataLifeCycle: 'Enabled'
-  LifecycleDataStorageClass: 'INTELLIGENT_TIERING'
+  LifecycleDataStorageClass: 'DEEP_ARCHIVE'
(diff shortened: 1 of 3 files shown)


brucehoff (approved):

brucehoff:
@zaro0508 Should the same change be applied to other existing SciComp buckets?  Will buckets created in the future in SciComp also have intelligent tiering configured?

zaro0508:
@brucehoff 
> Should the same change be applied to other existing SciComp buckets? 

Sure?  tell me which ones and i will add the parameters for those.

> Will buckets created in the future in SciComp also have intelligent tiering configured?

Not by default, it's up to the user to pass in the parameters to enable the bucket life cycle

brucehoff:
> tell me which ones and i will add the parameters for those.
I was thinking about the various buckets listed here:
https://github.com/Sage-Bionetworks/scicomp-provisioner/tree/master/config/prod

> Not by default, it's up to the user to pass in the parameters to enable the bucket life cycle
Would it be possible to make "intelligent tiering" the default?

zaro0508:
> I was thinking about the various buckets listed here: https://github.com/Sage-Bionetworks/scicomp-provisioner/tree/master/config/prod

not all of those are buckets but i guess you mean you want all buckets provisioned by the scicomp provisioner to be set to Intelligent tiering?   Remember, intelligent tierring cost extra $$ and may only be worth it if there is sufficient data in the bucket. 

> Would it be possible to make "intelligent tiering" the default?

Yes, this is possible but it's probably better to create a separate jira issue to implement that "behaviour" change.

brucehoff (approved):

PR #596, opened 2023-07-28
[IT-2300] Fix AWS security groups in Scicomp

zaro0508:
Only allow access to s3 sync instances from the sage VPN

### DIFF
--- a/config/prod/s3-synapse-sync.yaml
+++ b/config/prod/s3-synapse-sync.yaml
@@ -8,6 +8,7 @@ stack_tags:
 dependencies:
   - "prod/htan-synapse-sync-kms-key.yaml"
 parameters:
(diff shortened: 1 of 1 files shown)


config/prod/s3-synapse-sync.yaml
```diff
@@ -8,6 +8,7 @@ stack_tags:
 dependencies:
   - "prod/htan-synapse-sync-kms-key.yaml"
 parameters:
+  CIDR: "10.50.0.0/16"
```
jesusaurus, line 11: `10.50.0.0/16` is union station right? does this need to be `10.100.0.0/16` for vpn client access? i'm not entirely sure which to use

jesusaurus (approved):

PR #601, opened 2023-08-25
Update nf-syn23664726-s3.yaml

allaway:
Maybe this will resolve permission errors in Tower?

### DIFF
--- a/config/prod/nf-syn23664726-s3.yaml
+++ b/config/prod/nf-syn23664726-s3.yaml
@@ -21,7 +21,8 @@ parameters:
     - 'arn:aws:iam::694704239947:user/giri' ## Nebula Genomics access
     - 'arn:aws:iam::969434029702:user/audrisc' ##Stanford access (Audris Chiang)
     - 'arn:aws:iam::728882028485:role/ntap-add5-project-TowerForgeBatchHeadJobRole-X9Y068JVFXHC' ## Sage NF Tower access
-    - 'arn:aws:iam::728882028485:role/ntap-add5-project-TowerForgeBatchWorkJobRole-TYLG8H2QIS7R' ## Sage NF Tower access
+    - 'arn:aws:iam::728882028485:role/ntap-add5-project-TowerForgeBatchWorkJobRole-TYLG8H2QIS7R' ## Sage NF Tower access 
+    - 'arn:aws:iam::728882028485:role/ntap-add5-project-TowerForgeServiceRole-1KZP8QDA7AR8P'  ## Sage NF Tower access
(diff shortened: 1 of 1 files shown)


config/prod/nf-syn23664726-s3.yaml
```diff
@@ -21,7 +21,9 @@ parameters:
     - 'arn:aws:iam::694704239947:user/giri' ## Nebula Genomics access
     - 'arn:aws:iam::969434029702:user/audrisc' ##Stanford access (Audris Chiang)
     - 'arn:aws:iam::728882028485:role/ntap-add5-project-TowerForgeBatchHeadJobRole-X9Y068JVFXHC' ## Sage NF Tower access
-    - 'arn:aws:iam::728882028485:role/ntap-add5-project-TowerForgeBatchWorkJobRole-TYLG8H2QIS7R' ## Sage NF Tower access
+    - 'arn:aws:iam::728882028485:role/ntap-add5-project-TowerForgeBatchWorkJobRole-TYLG8H2QIS7R' ## Sage NF Tower access
```
zaro0508, line 24: there already a TowerForgeBatchHeadJobRole defined in the line above.  is this one intended to replace the one above?

allaway: Hi @zaro0508, there are three roles I am trying to add: TowerForgeBatchHeadJobRole, TowerForgeBatchWorkJobRole, TowerForgeServiceRole. 

 My end goal is to use Tower to index a non-Tower bucket (s3://nf-syn23664726-s3-bucket-n9uakf7bowwd/). Do you think that these three roles will be sufficient to do this? I tried just TowerForgeBatchHeadJobRole and TowerForgeBatchWorkJobRole and that didn't appear to work.

zaro0508: I'm sorry, I don't know Tower well enough to answer your question @allaway.  I would suggest asking @thomasyu888

thomasyu888: @zaro0508 I'm actually not certain.  @allaway is getting a permission issue, but these are the specific roles that are used when processing data on that tower workspace.  

Is the policy set up in a way to allow for indexing of files to a Synapse project via a different role?

allaway: After talking with Adam, I think I need to make the config change in this PR, but I also think I need to change the config for the nextflow project.

thomasyu888 (approved):
LGTM

zaro0508:
please fix the pre-commit failures @allaway then we can merge

allaway:
Oops, trailing whitespace. Fixed!

PR #606, opened 2023-09-08
[IT-2892] Security updates for s3-synapse-sync

jesusaurus:
Deploy dependabot security updates for the s3-synapse-sync lambda.

### DIFF
--- a/config/prod/s3-synapse-sync.yaml
+++ b/config/prod/s3-synapse-sync.yaml
@@ -1,6 +1,6 @@
 template:
   type: "http"
-  url:  "https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/s3-synapse-sync/v1.6.1/s3-synapse-sync.yaml"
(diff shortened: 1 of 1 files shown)


config/prod/s3-synapse-sync.yaml
```diff
@@ -1,6 +1,6 @@
 template:
   type: "http"
-  url:  "https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/s3-synapse-sync/v1.6.1/s3-synapse-sync.yaml"
+  url:  "https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/s3-synapse-sync/v1.7.2/s3-synapse-sync.yaml"
```
xschildw, line 3: Should it be 1.7.2 (from tags in project)?

jesusaurus: oh whoops, good catch, I hadn't noticed that the tag format changed

PR #642, opened 2024-09-22
DYP access to HTAN AWS bucket requrst.

PozhidayevaDarya:
Requesting to be added to relevant HTAN buckets in AWS.


adamjtaylor:
Approving this request as HTAN DCC PI at Sage. @PozhidayevaDarya is our collaborator at ISB and requires access.

zaro0508 (approved):

PR #646, opened 2024-10-18
Update deprecated program codes

zaro0508:
The cost center codes that were in the deprecation list are actually
valid.  The codes matach up to codes from finance, the description
portion was not exactly the same.  Update to match the ones from
finance exactly.


brucehoff (approved):
I just spot checked a few of the updated tags against
[this](https://sagebionetworks.jira.com/wiki/spaces/IT/pages/2553544733/Finance+Program+Codes+for+AWS+tagging) list.

ConsoleCatzirl (approved):

PR #648, opened 2024-10-21
[IT-3953] disable SameRegionResourceAccessToBucket

zaro0508:
issue https://sagebionetworks.jira.com/browse/IT-3953 caused the AWS deployment to fail so we need to temporarily disable SameRegionResourceAccessToBucket

### DIFF
--- a/config/prod/TESLA-download-bucket.yaml
+++ b/config/prod/TESLA-download-bucket.yaml
@@ -22,7 +22,7 @@ parameters:
   # Data transfers within the same region between AWS resources are free.
   # true to restrict downloading files from this bucket to only AWS resources (e.g. EC2 , Lambda) within the same region as this bucket.
   # This will not allow even the owner of the bucket to download objects in this bucket when not using an AWS resource in the same region!
-  SameRegionResourceAccessToBucket: "true"
+#  SameRegionResourceAccessToBucket: "true"
 
   # (Optional) Allow accounts, groups, and users to access bucket. (default is no access).
   GrantAccess:
--- a/config/prod/amp-mayo-sinai.yaml
+++ b/config/prod/amp-mayo-sinai.yaml
@@ -29,7 +29,7 @@ parameters:
   # Restrict downloading files from this bucket to only AWS resources (e.g. EC2 , Lambda)
   # within the same region as this bucket.  This will not allow even the owner of the bucket
   # to download objects in this bucket when not using an AWS resource in the same region!
-  SameRegionResourceAccessToBucket: "true"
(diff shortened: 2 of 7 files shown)


brucehoff (commented):
The error you put in IT-3953 says, "failed to create RestrictBucketDownloadRegionFunction".  If the currently existing buckets already have the IPAddressRestictionLambda, why are they affected by a failure to create the Lambda?  Shouldn't the problem affect only new buckets?

If we merge this PR then seven buckets will lose their protection against costly egress. How is this risk mitigated?

zaro0508:
> The error you put in IT-3953 says, "failed to create RestrictBucketDownloadRegionFunction". If the currently existing buckets already have the IPAddressRestictionLambda, why are they affected by a failure to create the Lambda? Shouldn't the problem affect only new buckets?

It's affecting existing buckets because there was a [change to update tags](https://github.com/Sage-Bionetworks/scicomp-provisioner/pull/646) on existing buckets.  

> If we merge this PR then seven buckets will lose their protection against costly egress. How is this risk mitigated?

There are PRs in this repo waiting to merge so my idea was to temporarily disable RestrictBucketDownloadRegionFunction to keep deployments working.  The alternative is to fix IT-3953 instead of temporarily disabling RestrictBucketDownloadRegionFunction.  Another alternative is to manually deploy PRs until we fix IT-3953.  Is there an option that you prefer @brucehoff ?

brucehoff:
> > The error you put in IT-3953 says, "failed to create RestrictBucketDownloadRegionFunction". If the currently existing buckets already have the IPAddressRestictionLambda, why are they affected by a failure to create the Lambda? Shouldn't the problem affect only new buckets?
> 
> It's affecting existing buckets because there was a [change to update tags](https://github.com/Sage-Bionetworks/scicomp-provisioner/pull/646) on existing buckets.
> 
> > If we merge this PR then seven buckets will lose their protection against costly egress. How is this risk mitigated?
> 
> There are PRs in this repo waiting to merge so my idea was to temporarily disable RestrictBucketDownloadRegionFunction to keep deployments working. The alternative is to fix IT-3953 instead of temporarily disabling RestrictBucketDownloadRegionFunction. Another alternative is to manually deploy PRs until we fix IT-3953. Is there an option that you prefer @brucehoff ?

Thank you for presenting these options. As suggested in our meeting today, can you please prioritize fixing IT-3953?

zaro0508:
replaced with PRs  https://github.com/Sage-Bionetworks/aws-infra/pull/416 and https://github.com/Sage-Bionetworks/scicomp-provisioner/pull/649

PR #665, opened 2025-09-19
[IT-4556] Add S3 inventory to HTAN buckets

ConsoleCatzirl:
Update to version 1.7.6 of `s3-synapse-sync-bucket.j2`.

This version adds a second bucket with `-inventory` appended
to the name, and configures the origin bucket to write S3 Inventory
into the new bucket.

Depends on https://github.com/Sage-Bionetworks/s3-synapse-sync/pull/63


brucehoff:
> and written back into the same bucket with an inventory/ prefix.

I wonder if there is any chance of mixing the inventory files in with the bucket's data files. E.g., if the bucket is configured to automatically index its content into Synapse will the inventory start appearing there?

ConsoleCatzirl:
Indexing the inventory data into Synapse would be a feature, but writing the inventory into a dedicated bucket allows us to retain the inventory data beyond the source bucket lifecycle expiration, as well as beyond the deletion of the source bucket.

zaro0508 (approved):

